PR #33, opened 2020-03-26
Quote the "$@" so that quoted params are correctly passed through

staer:


staer:
Verified to work with a cluster with spaces in the name

PR #37, opened 2020-07-06
RHEL images would not get appropriate attributes set

staer:


staer:
Closing because it's fixed in a newer PR

PR #38, opened 2020-07-09
add parameter for maxPrice

mvrequa:


jermth:
I tested this with changing max prices, seems to work. Awesome.

PR #69, opened 2021-11-05
Feature/2.5.0

ryanhamel:
* Hostnames now match nodenames by default.
* create and remove_nodes now take a list of relevant nodes.
* 'reconfigure' also added as a way to do all of the actions of 'scale' without regenerating the nodes in CycleCloud.
* /opt/cycle/slurm/slurm_healthcheck.py added to each execute node, adding the ability to safely run health check scripts when no jobs are running. If the health check fails, the node will be left in a 'drained' state until it is shutdown.
* Allow users to specify arbitrary slurm.conf entries to include in the default template
* Allow users to opt out of installing Slurm packages when they've already installed it in their custom image
* Adjust the build script to optionally include PMIX if it's been installed on the image used for building Slurm


tbugfinder:
Hi @ryanhamel ,
I see there's a 2.5.1 release pending as a bugfix.
Is there any plan to release 2.5.1 soon or is 2.5.0 ready to use?

PR #78, opened 2022-02-04
Fix bug with uppercase cluster names causing a mismatch

anhoward:


bwatrous:
@anhoward  would you mind checking if this PR is still required and updating it if so?

Issue #80, opened 2022-02-10
login nodes can't run accounting commands


themorey:
the default slurm.conf uses `AccountingStorageHost="localhost"` instead of the resolvable name of the slurmdbd host (ie. `scheduler`)

themorey:
@anhoward

PR #86, opened 2022-03-04
update accounting configs for HA

themorey:
- if accounting is enabled 
    -- set `jobacct_gather` to `linux`
    -- set hostname of scheduler instead of using localhost (for login node support)
- if HA is enabled add backup scheduler as AccountingStorageBackupHost


themorey:
issue #80

themorey:
@anhoward I can't add a reviewer so you get this 😀

PR #89, opened 2022-03-10
slurm.use_pcpu changes for HT VM

themorey:
remove `slurm.use_pcpu=false` from template htc nodearray and place it as default setting in `default.rb`


themorey:
@anhoward

PR #104, opened 2022-11-06
Update default instance type

atomic-penguin:
Change D12_v2 to D16_v4, better default price / performance value and supports hyperv gen2.

gen 2 support matrix: https://learn.microsoft.com/en-us/azure/virtual-machines/generation-2

### DIFF
--- a/templates/slurm-beegfs.txt
+++ b/templates/slurm-beegfs.txt
@@ -110,7 +110,7 @@ Order = 10
         Label = Scheduler VM Type
         Description = The VM type for scheduler node
         ParameterType = Cloud.MachineType
-        DefaultValue = Standard_D12_v2
+        DefaultValue = Standard_E4_v4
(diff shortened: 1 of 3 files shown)


templates/slurm-beegfs.txt
```diff
@@ -110,7 +110,7 @@ Order = 10
         Label = Scheduler VM Type
         Description = The VM type for scheduler node
         ParameterType = Cloud.MachineType
-        DefaultValue = Standard_D12_v2
+        DefaultValue = Standard_D16_v4
```
dougclayton, line 113: The D12 is 4 cores/28 GB for $.368/hr. The D16_v4 is 16 cores/64 GB for $.896/hr. Seems like the D4_v4 (4/16 for $.224/hr) would be better, or at least the D8 in the middle.

PR #110, opened 2023-02-24
Slurm accounting with cost reporting

aditigaur4:
Adds cost reporting to slurm accounting and creates a new `azslurm cost` CLI to append costs to slurm accounting data.

Also adds a prolog slurmctld script that records vm size related information such as sku, region, priority, cpu/gpu counts etc and stores them in admincomment field in slurm, so that cost reporting can fetch relevant cost data.

### DIFF
--- a/slurm/src/slurmcc/cli.py
+++ b/slurm/src/slurmcc/cli.py
@@ -150,7 +150,7 @@ class SlurmCLI(CommonCLI):
         to SLURM Job Accounting data. This is an experimental
         feature.
         """
-        config['cache_root'] = "/tmp"
+
         curr = datetime.today()
         delta = timedelta(days=365)
 
@@ -352,6 +352,7 @@ class SlurmCLI(CommonCLI):
         parser.add_argument(
             "--accounting-subscription-id", dest="accounting__subscription_id"
         )
+        parser.add_argument("--cost-cache-root", dest="cost__cache_root")
 
     def _initconfig(self, config: Dict) -> None:
         # TODO
--- a/slurm/src/slurmcc/cli.py
+++ b/slurm/src/slurmcc/cli.py
@@ -11,7 +11,7 @@ import sys
 import traceback
 import time
 from argparse import ArgumentParser
-from datetime import date,datetime,timedelta
+from datetime import date,datetime,timedelta,time
 from math import ceil
 from typing import Any, Callable, Dict, Iterable, List, Optional, Set, TextIO, Union
 
--- a/slurm/src/slurmcc/cli.py
+++ b/slurm/src/slurmcc/cli.py
@@ -11,6 +11,7 @@ import sys
 import traceback
 import time
 from argparse import ArgumentParser
+from datetime import date,datetime,timedelta
 from math import ceil
 from typing import Any, Callable, Dict, Iterable, List, Optional, Set, TextIO, Union
 
@@ -22,6 +23,7 @@ from hpc.autoscale.clilib import (
     main as clilibmain,
 )
 from hpc.autoscale import hpctypes as ht
+from hpc.autoscale.cost.azurecost import azurecost
 from hpc.autoscale.hpctypes import Memory
 from hpc.autoscale import util as hpcutil
 from hpc.autoscale.job.demandprinter import OutputFormat
@@ -36,6 +38,7 @@ from hpc.autoscale.job.schedulernode import SchedulerNode
 from . import partition as partitionlib
 from . import util as slutil
 from . import CyclecloudSlurmError
+from . import cost
 from hpc.autoscale.results import AllocationResult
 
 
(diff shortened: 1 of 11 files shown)


slurm/src/slurmcc/cli.py
```diff
@@ -129,6 +132,45 @@ def _slurm_node_name_completer(
...
+        #parser.add_argument("--user", action='store_true', help="Show costs aggregated by user")
+        #parser.add_argument("-f", "--fmt", type=str, help="Comma separated list of formatting options")
+
+    def cost(self, config: Dict, start, end, out):
+        """
+        Cost analysis and reporting tool that maps Azure costs
+        to SLURM Job Accounting data. This is an experimental
+        feature.
+        """
+        config['cache_root'] = "/tmp"
```
ryanhamel, line 153: 1) What if the cx wants to override this? 
2) If it can be changed, it should be under a different namespace - like config.get("cost", {}).get("cache_root", "/tmp")

slurm/src/slurmcc/cost.py
```diff
@@ -0,0 +1,256 @@
...
+    except Exception as e:
+        log.error(e)
+        raise
+    log.debug(f"output: {output.stdout}")
+    return output
+
+def get_sacct_fields():
+
+    options = []
+    cmd = "/usr/bin/sacct -e"
```
ryanhamel, line 36: just as an example of the the run_command(str) vs run_command(list[str])
you would do run_command(["/usr/sbin/sacct", "-e"]) instead.

slurm/src/slurmcc/cost.py
```diff
@@ -0,0 +1,256 @@
...
+    out = run_command(cmd)
+    for line in out.stdout.split('\n'):
```
ryanhamel, line 38: use splitlines() instead

slurm/src/slurmcc/cost.py
```diff
@@ -0,0 +1,256 @@
...
+class CostSlurm:
+    def __init__(self, start:str, end: str, cluster: str, fmt: str=None) -> None:
+
+        self.start = start
+        self. end = end
+        self.cluster = cluster
+        self.sacct = "/usr/bin/sacct"
+        self.squeue = "/usr/bin/squeue"
+        self.sacctmgr = "/usr/bin/sacctmgr"
+        cache_root = "/tmp"
```
ryanhamel, line 74: If we are making this configurable, this should get injected via init args. It may be worth it to pass in the config, because it is quite possible other things hard coded here may need to be overridden - /usr/bin/* etc, though another option is to just use the path returned from which.which() and let them setup their path themselves.

slurm/src/slurmcc/cost.py
```diff
@@ -0,0 +1,256 @@
...
+        self.stats = Statistics()
+        self.cache = f"{cache_root}/slurm"
+        try:
+            os.makedirs(self.cache, 0o777, exist_ok=True)
+        except OSError as e:
+            log.error("Unable to create cache directory {self.cache}")
```
ryanhamel, line 80: You are missing the leading f here on the format string.

slurm/src/slurmcc/cost.py
```diff
@@ -0,0 +1,256 @@
...
+        cmd = self._construct_command()
+        with open(_job_rec_file, 'w') as fp:
+            output = run_command(cmd, stdout=fp)
+            if output.returncode:
+                log.error("Could not fetch slurm records")
+        return _job_rec_file
+
+    def parse_admincomment(self, comment: str):
+
+        return json.loads(comment)
```
ryanhamel, line 161: I recommend using a helper function here. You will definitely get malformed json eventually, and you want to give as much info to the user on where this is happening. So do something that tries to parse it, logs the invalid comment, and then raises an exception. Anyone who calls this func can then capture that error and log which specific comment caused this.

slurm/src/slurmcc/cost.py
```diff
@@ -0,0 +1,256 @@
...
+                comment_d = self.parse_admincomment(admincomment)[0]
+                sku_name = comment_d['vm_size']
+                cpupernode = comment_d['pcpu_count']
+                region = comment_d['location']
+                spot = comment_d['spot']
+            except json.JSONDecodeError as e:
+                log.debug(f"Cannot parse admincomment job={row.jobid} cluster={row.cluster}")
+                self.stats.admincomment_err += 1
+                self.stats.unprocessed += 1
+                continue
```
ryanhamel, line 193: You might want to add - what if it is valid json but it is just missing the fields you need?
except KeyError as ke:
...

slurm/src/slurmcc/cost.py
```diff
@@ -0,0 +1,256 @@
...
+            charge_factor = float(row.ncpus) / float(cpupernode)
+
+            az_fmt = azcost.get_job(sku_name, region, spot)
+            charged_cost = ((az_fmt.rate/3600) * float(row.elapsedraw)) * charge_factor
+            c_fmt = self.c_fmt_t(cost=charged_cost)
+            if (region,sku_name) not in self.stats.cost_per_sku:
```
ryanhamel, line 199: TOTALLY OPTIONAL - this is more of how we can use typing to prevent future changes from causing issues.

if you are using a tuple as a key here, it is very easy to flip these down the line. There are two ways to prevent this:
region: Region
sku_name: SkuName
where 
Region = typing.NewType("Region", str)
SkuName = typing.NewType("SkuName", str)
CostPerSkuKey = typing.Newtype("CostPerSkuKey", Tuple[Region, SkuName])
and when declaring costs_per_sku
costs_per_sku: Dict[CostPerSkuKey, float]

This means you need to propagate Region, SkuName type everywhere. This is probably what I would do, but it is not the only way.

Another way is to just declare the key type like this - a simple Tuple[str, str]
CostPerSkuKey = typing.Newtype("CostPerSkuKey", Tuple[str, str])
and use a function to create the key. Note I forced it to use "key word args", like func(key=value)
so you have to write key = new_cost_per_sku_key(region=region, sku_name=sku_name)
def new_cost_per_sku_key(**kwargs) -> CostPerSkuKey:
     region = kwargs.pop("region")
     sku_name = kwargs.pop("sku_name")
     assert not kwargs, f"unexpected arguments {kwargs}"
     return CostPerSkuKey(region, sku_name)

slurm/src/slurmcc/cost.py
aditigaur4: Fixed

slurm/src/slurmcc/cost.py
aditigaur4: fixed

slurm/src/slurmcc/cost.py
aditigaur4: Done, its now in init args

slurm/src/slurmcc/cost.py
aditigaur4: Done

slurm/src/slurmcc/cost.py
aditigaur4: Done

slurm/src/slurmcc/cost.py
```diff
@@ -0,0 +1,256 @@
...
+from collections import namedtuple
+from hpc.autoscale.cost.azurecost import azurecost
+
+log = logging.getLogger('cost')
+
+def run_command(cmd: str, stdout=subprocess.PIPE, stderr=subprocess.PIPE):
+    """
+    run arbitrary command
+    """
+    cmd_list = cmd.split(' ')
```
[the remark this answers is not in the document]
aditigaur4: Done

slurm/src/slurmcc/cli.py
aditigaur4: Changed, this now comes from initconfig. If not present only then it defaults to /tmp

slurm/src/slurmcc/cost.py
aditigaur4: As discussed, not doing this one for now

PR #111, opened 2023-03-20
Feature/acct 3x pr1

ryanhamel:


docker-rpmbuild.sh
```diff
@@ -1,19 +1,28 @@
 #!/bin/bash
+set -e
+cd $(dirname $0)/install
```
aditigaur4, line 3: I was getting an error here that the install directory does not exist, and in this version it does not exist in the root of the repository. There is an install directory in the `slurm` dir, but since we cd into it, the rest of the script does not work (there is a `pwd` afterwards). So to get it to work i had to comment out line 2 to 6. Either there is a bug or i am not sure how to invoke the script. I ran it like : `docker-rpmbuild.sh` from the root of the repo. I tried running it from `slurm` directory as well, but as mentioned above, the `pwd` later on in line 18 fails.

docker-rpmbuild.sh
```diff
@@ -1,19 +1,28 @@
...
-SLURM20="20.11.9"
+TMP_BINS=slurm-pkgs-tmp
+rm -rf $TMP_BINS
+mkdir $TMP_BINS
 
-rm -rf blobs-build-tmp
-mkdir blobs-build-tmp
 if command -v docker; then
+  extra_args=
+  if command -v podman; then
```
aditigaur4, line 14: should this be `elif` instead? but since this is inside the outer if loop i am not sure what the purpose is here. If docker is not installed in the system, this won't work anyways. for reference i used the script with podman, and I had to comment out parts of it it to get it to work. We could do something like `if command -v podman; then $runtime=podman` and then use $runtime variable in line 18,19,20. upto you though!

docker-rpmbuild.sh
```diff
@@ -1,19 +1,28 @@
...
-  docker run -v $(pwd)/specs/default/cluster-init/files:/source -v $(pwd)/blobs-debian-tmp:/root/rpmbuild/RPMS/x86_64 -ti ubuntu:18.04 /bin/bash /source/01-build-debs.sh
-  # we only want the rpms for almalinux8 (i.e. slurm22) but we want all of the debs
-  mv blobs-debian-tmp/*.deb blobs/
-  mv blobs-debian-tmp/*$SLURM22*.rpm blobs/
-  rm -rf blobs-debian-tmp
+  docker run -v $(pwd)/specs/default/cluster-init/files:/source -v $(pwd)/$TMP_BINS:/root/rpmbuild/RPMS/x86_64 $extra_args -ti almalinux:8.5 /bin/bash /source/00-build-slurm.sh centos
+  docker run -v $(pwd)/specs/default/cluster-init/files:/source -v $(pwd)/$TMP_BINS:/root/rpmbuild/RPMS/x86_64 -ti centos:7 /bin/bash /source/00-build-slurm.sh centos
+  docker run -v $(pwd)/specs/default/cluster-init/files:/source -v $(pwd)/$TMP_BINS:/root/rpmbuild/RPMS/x86_64 $extra_args -ti ubuntu:18.04 /bin/bash /source/01-build-debs.sh
+  
+  mv $TMP_BINS/* slurm-pkgs/
```
aditigaur4, line 22: This line failed for me: since slurm-pkgs directory was not there already. I think it expects the directory to be pre-made.

`mv: target 'slurm-pkgs/' is not a directory`

docker-rpmbuild.sh
```diff
@@ -1,19 +1,28 @@
...
+  extra_args=
+  if command -v podman; then
+    extra_args="--privileged"
+  fi
 
-  docker run -v $(pwd)/specs/default/cluster-init/files:/source -v $(pwd)/blobs-debian-tmp:/root/rpmbuild/RPMS/x86_64 -ti almalinux:8.5 /bin/bash /source/00-build-slurm.sh centos $SLURM22
-  docker run -v $(pwd)/specs/default/cluster-init/files:/source -v $(pwd)/blobs-debian-tmp:/root/rpmbuild/RPMS/x86_64 -ti centos:7 /bin/bash /source/00-build-slurm.sh centos $SLURM20
-  docker run -v $(pwd)/specs/default/cluster-init/files:/source -v $(pwd)/blobs-debian-tmp:/root/rpmbuild/RPMS/x86_64 -ti ubuntu:18.04 /bin/bash /source/01-build-debs.sh
-  # we only want the rpms for almalinux8 (i.e. slurm22) but we want all of the debs
-  mv blobs-debian-tmp/*.deb blobs/
```
aditigaur4, line 14: I believe this line can be used down below. right now the build.sh script at the root expects slurm rpms to be in blobs directory for packaging the azure-slurm-pkg. So we should try and copy the rpms in this script. This line can be altered like this: `cp $TMP-BINS/* blobs/`

docker-rpmbuild.sh
ryanhamel: Agreed and fixed with latest push

slurm/install/rhel.sh
```diff
@@ -1,9 +1,26 @@
...
 else
-    yum  -y install $(ls blobs/*.rpm | grep -v slurmdbd)
-fi
\ No newline at end of file
+    yum -y install python3
+fi
+
+
+if [ ${SLURM_ROLE} == "scheduler" ]; then
+    yum  -y install $(ls slurm-pkgs/*${SLURM_VERSION}.el${OS_VERSION}*.rpm)
```
aditigaur4, line 21: Encountering errors in this line.

I believe in install.py when we call this script, it may not necessarily use the same working directory as the cluster-init, (from which install.py is called). 

The error i am getting is this:

`ls: cannot access 'slurm-pkgs/*23.02.0-1.el8.6*.rpm': No such file or directory
usage: yum install [-c [config file]] [-q] [-v] [--version]
                   [--installroot [path]] [--nodocs] [--noplugins]
                   [--enableplugin [plugin]] [--disableplugin [plugin]]
                   [--releasever RELEASEVER] [--setopt SETOPTS]
                   [--skip-broken] [-h] [--allowerasing] [-b | --nobest] [-C]
                   [-R [minutes]] [-d [debug level]] [--debugsolver]
                   [--showduplicates] [-e ERRORLEVEL] [--obsoletes]
                   [--rpmverbosity [debug level name]] [-y] [--assumeno]
                   [--enablerepo [repo]] [--disablerepo [repo] | --repo
                   [repo]] [--enable | --disable] [-x [package]]
                   [--disableexcludes [repo]] [--repofrompath [repo,path]]
                   [--noautoremove] [--nogpgcheck] [--color COLOR] [--refresh]
                   [-4] [-6] [--destdir DESTDIR] [--downloadonly]
                   [--comment COMMENT] [--bugfix] [--enhancement]
                   [--newpackage] [--security] [--advisory ADVISORY]
                   [--bz BUGZILLA] [--cve CVES]
                   [--sec-severity {Critical,Important,Moderate,Low}]
                   [--forcearch ARCH]
                   PACKAGE [PACKAGE ...]
yum install: error: the following arguments are required: PACKAGE
Traceback (most recent call last):
  File "install.py", line 419, in <module>
    main()
  File "install.py", line 396, in main
    run_installer(settings, os.path.abspath(f"{args.platform}.sh"), args.mode)
  File "install.py", line 89, in run_installer
    subprocess.check_call([path, mode, s.slurmver])
  File "/opt/cycle/jetpack/system/embedded/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['/opt/cycle/jetpack/system/bootstrap/azure-slurm-install/rhel.sh', 'scheduler', '23.02.0-1']' returned non-zero exit status 2.`

slurm/install/install.py
```diff
@@ -429,7 +393,7 @@ def main() -> None:
     munge_key(settings)
 
     # runs either rhel.sh or ubuntu.sh to install the packages
-    run_installer(os.path.abspath(f"{args.platform}.sh"), args.mode)
+    run_installer(settings, os.path.abspath(f"{args.platform}.sh"), args.mode)
```
aditigaur4, line 396: This is problematic as script is expecting a relative path to slurm-pkgs. although in cluster-init we are cd ing to the directory, subprocess is running it from absolute path.

slurm/install/install.py
```diff
@@ -86,8 +85,8 @@ def setup_users(s: InstallSettings) -> None:
     )
 
 
-def run_installer(path: str, mode: str) -> None:
-    subprocess.check_call([path, mode])
+def run_installer(s: InstallSettings, path: str, mode: str) -> None:
```
aditigaur4, line 88: I overall think we need to be a bit more careful in scripts where we are using relative/absolute paths. right now there are only 2 scripts, but if we plan to add more, then we need to make it look more ansible like. right now shell to python then to back shell is getting a bit convoluted. We can make immediate fixes but something to keep in mind.

slurm/install/rhel.sh
aditigaur4: for reference, the rpms are there:

[root@slurm301-1-scheduler azure-slurm-install]# pwd
/opt/cycle/jetpack/system/bootstrap/azure-slurm-install
[root@slurm301-1-scheduler azure-slurm-install]# ls slurm-pkgs | grep 23.02
slurm_23.02.0-1_amd64.deb
slurm-23.02.0-1.el7.x86_64.rpm
slurm-23.02.0-1.el8.x86_64.rpm
slurm-devel_23.02.0-1_amd64.deb
slurm-devel-23.02.0-1.el7.x86_64.rpm
slurm-devel-23.02.0-1.el8.x86_64.rpm

(havent pasted the full output)

ryanhamel: This issue is because OS_VERSION is coming in at 8.6 instead of 8 - I'll push a fix.

slurm/install/install.py
ryanhamel: Good point. For the actual scripts they are running, I added a cd $(realpath $(dirname $0)) at the root so that it is irrelevant from where it is invoked for now.

sbin/suspend_program.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+node_list=$(echo $@ | sed "s/ /,/g")
```
aditigaur4, line 2: Unfortunately this has not fixed the issue i was seeing earlier. I am still seeing the same issue.

aditigaur4: Mysteriously the problem only comes in htc partition and not hpc 🤔

PR #130, opened 2023-07-18
Add warning about shared disk and clean up template

dougclayton:


aditigaur4:
Is the intention that this needs to go into slurm 3.0.2 release? we are not unfortunately using master branch for that.

dougclayton:
Yes, that was the intention. Can you cherry-pick the commit to the right branch? Also I did the same PR against the PBS master branch...is that wrong too?

———
Doug Clayton
Software Developer, Azure HPC
________________________________

aditigaur4:
Yep will do

PR #133, opened 2023-08-02
Fix login nodes and Remove unused chef recipes

aditigaur4:
Run slurm installation on login nodes, so slurm binaries are accessible from login nodes. Allows users to run `squeue`, `srun` etc from the logins.


SkylinSky (approved):
Seems like a clean way to separate installation paths for login and execute nodes

PR #134, opened 2023-08-02
Fix SSL certificates used for accounting

aditigaur4:
Azure Maria DB now requires combined SSL client certificates from 2 authorities. . This PR fixes the client certificates used by Azure Maria DB, and also adds a new certificate_url option for user to specify the certificate. If a user has specified certificate then it is used.

 The issue is described here: https://learn.microsoft.com/en-us/azure/mariadb/concepts-certificate-rotation?WT.mc_id=Portal-SqlAzureExtension

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -19,6 +19,7 @@ if [ ! -e blobs ]; then
     mkdir blobs
 fi
 
+wget -k -O slurm/install/AzureCA.pem  https://github.com/Azure/cyclecloud-slurm/releases/download/2.7.3/AzureCA.pem
 # ls slurm/install/slurm-pkgs/*.rpm > /dev/null || (echo you need to run docker-rpmbuild.sh first; exit 1)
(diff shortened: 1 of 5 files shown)


aditigaur4:
I need to make one more change on this.

SkylinSky (approved):
I like using copy_file's parameter to set the permissions instead of calling chown. 
I have a feeling   _add("suse.sh", "suse.sh", 600) might be a part of a different change (SuSe support perhaps?), but it seems minor enough

aditigaur4:
> I like using copy_file's parameter to set the permissions instead of calling chown. I have a feeling _add("suse.sh", "suse.sh", 600) might be a part of a different change (SuSe support perhaps?), but it seems minor enough

Thats a good catch, the suse change should not be going in this commit. I ll fix this.

ryanhamel (approved):

PR #138, opened 2023-08-04
Feature/remove automatic dynamic nodes

ryanhamel:
Remove automatic dynamic nodes, relying only on pre-created nodes via scontrol create commands.


slurm/src/slurmcc/allocation.py
```diff
@@ -24,67 +24,20 @@ def resume(
     partitions: List[partitionlib.Partition],
 ) -> BootupResult:
     name_to_partition = {}
-    feature_to_partition = {}
```
aditigaur4, line 27: This change does not have to be undone, we still need custom feature flags.

aditigaur4 (dismissed):

aditigaur4 (approved):

PR #139, opened 2023-08-04
Bugfix/fix gres conf

ryanhamel:
Fixes write location for gres.conf during azslurm scale, as well useful warnings if /etc/slurm/gres.conf is not a symlink or how to create the missing symlink.

### DIFF
--- a/slurm/src/slurmcc/cli.py
+++ b/slurm/src/slurmcc/cli.py
@@ -422,7 +422,17 @@ class SlurmCLI(CommonCLI):
         os.makedirs(backup_dir)
 
         azure_conf = os.path.join(sched_dir, "azure.conf")
-        gres_conf = os.path.join(slurm_conf_dir, "gres.conf")
+        gres_conf = os.path.join(sched_dir, "gres.conf")
+        linked_gres_conf = os.path.join(slurm_conf_dir, "gres.conf")
+        if os.path.isfile(linked_gres_conf) and not os.path.islink(linked_gres_conf):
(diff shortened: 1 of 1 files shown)


aditigaur4 (changes requested):
This PR only needs to contain gres.conf changes. Please try to do your work as separate commits to avoid this issue.

aditigaur4 (commented):
When i try to generate `gres.conf ` from `azslurm scale` I get this:

```
[root@slurm302A-scheduler ~]# azslurm scale
WARNING: please run 'ln -s /sched/gres.conf /etc/slurm/gres.conf' && chown slurm:slurm /etc/slurm/gres.conf
```

Why can't we just add this here? and generate /etc/slurm/gres.conf in code?

ryanhamel:
> When i try to generate `gres.conf ` from `azslurm scale` I get this:
> 
> ```
> [root@slurm302A-scheduler ~]# azslurm scale
> WARNING: please run 'ln -s /sched/gres.conf /etc/slurm/gres.conf' && chown slurm:slurm /etc/slurm/gres.conf
> ```
> 
> Why can't we just add this here? and generate /etc/slurm/gres.conf in code?

As a rule, azlsurm is not supposed to be touching anything under /etc/slurm - it only recreates things in /sched/ - that way if a CX decides to use a completely different self-managed gres.conf, for example, they can handle it this way.

We set up the link properly during installation, but the goal of azslurm scale is not to repair those links so that a CX can manage those configuration files themselves. I did want to add a warning about how to repair it, at least though.

aditigaur4 (approved):

PR #144, opened 2023-08-05
Do not restart slurmctld for configuration updates.

aditigaur4:
azslurm scale is used for updating dynamic nodes and other slurm configurations. But not all changes (especially dynamic nodes) require restarting slurmctld. Restarting slurmctld should be left upto the user. Doing it on behalf of the user especially for a use case that is perfectly supported without having to restart slurmctld, can be quite disruptive. For other use cases, we should just print a warning message.

### DIFF
--- a/slurm/src/slurmcc/cli.py
+++ b/slurm/src/slurmcc/cli.py
@@ -428,7 +428,7 @@ class SlurmCLI(CommonCLI):
             msg = f"{linked_gres_conf} should be a symlink to {gres_conf}! Changes will not take effect locally."
             print("WARNING: " + msg, file=sys.stderr)
             logging.warning(msg)
-        
+
         if not os.path.exists(linked_gres_conf):
             msg = f"please run 'ln -s {gres_conf} {linked_gres_conf}' && chown slurm:slurm {linked_gres_conf}"
             print("WARNING: " + msg, file=sys.stderr)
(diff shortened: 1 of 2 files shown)


slurm/src/slurmcc/cli.py
```diff
@@ -458,9 +458,8 @@ def scale(
             _generate_gres_conf(partition_dict, fw)
         shutil.move(gres_conf + ".tmp", gres_conf)
 
-        logging.info("Restarting slurmctld...")
-        check_output(["systemctl", "restart", "slurmctld"])
-
+        msg = f"Warning: For changes impacting non-dynamic partitions, please run `systemctl restart slurmctld` for changes to take effect."
```
bwatrous, line 461: Agreed that restarting slurmctld can be disruptive...   
But this is also a breaking behavior change to existing users that may script "azslurm scale".   
Perhaps we should do something like add a "--restartSlurmctld false" option on the command line to get the new behavior (with default true)?

aditigaur4: I agree it could be breaking for users who may have scripted against this, I agree that an option can be added, but how about we make the default equals false? I mean user wont see any issue in their scripting, all that would happen that slurmctld wont be restarted in the background. of all the use cases i have seen only changing node counts on static partitions (hpc and htc) is the change that requires an actual slurmctld restart. But because of dynamic partitions, people are having to run `azslurm scale` more often (by changing vm sizes etc), and slurmctld does not need to be restarted for that.

anhoward: Could we add the "--restartSlurmctld=true" option to the azslurm command? I do agree that restarting by default is probably not the best option, even for users who may have scripted against it (honestly if they've scripted their scale commands and are doing things expecting a restart, they're probably doing something odd anyway). But adding the option does provide an easy way to do the change and restart all in one go.

aditigaur4:
no longer needed

PR #150, opened 2023-08-08
For dynamic partition make config.MultiSelect true by default.

aditigaur4:
For dynamic partitions having the default of allowing multiple VM sizes as default makes them more useful.


AnastasiiaEvdokimova (commented):
It seems like config is capitalized in the other parameters. Is there something that makes this one special?

AnastasiiaEvdokimova (approved):

PR #154, opened 2023-08-16
Set cuser.base_home_dir default to /shared/home as a workaround

### LINKED ISSUE #153: Some users have cuser.base_home_dir set to /home and not /shared/home
Some users are having to specify cuser.base_home_dir=/shared/home in their template. While this does not specifically seem to be related to AzSlurm, we can easily mitigate this going forward.

ryanhamel:
A workaround for something some users are experiencing - cuser.base_home_dir is set to /home instead of /shared/home in some environments.

Fixes #153


aditigaur4 (approved):

themorey:
bumping this for 8.4.1 and Slurm 3.0.3...my login node using `microsoft-dsvm:ubuntu-hpc:2004:latest` had its home dir set to `/home` instead of `/shared/home` like the rest of the cluster.  Applying the workaround fixed it.

PR #155, opened 2023-08-16
Add build improvements.

aditigaur4:
- Move all scripts (such as build scripts) to a util directory
- Use a dockerfile that builds an image for packaging azslurm, so that build environment gets cached locally.
- Make sure slurm_supported_versions.py points to a version that is downloadable schedmd.slurm.com. Because the versions are appended with a "-1", when download url links are created, we create the wrong link and docker_rpmbuild.sh fails.
- make all scripts use podman or docker depending on which is available.
Signed-off-by: Aditi Gaur <aditigaur@microsoft.com>


ryanhamel (approved):
Looks great.

PR #164, opened 2023-08-29
Add HA support

ryanhamel:
Wait for secondary hostname to be reported to CC before adding to slurm.conf Handle setting the hostname of scheduler-ha nodes
add installlib.append_file to safely append files with commented out hashes


slurm/install/install.py
```diff
@@ -257,11 +287,28 @@ def accounting(s: InstallSettings) -> None:
...
+    """
+    Only the primary scheduler should be creating files under /sched.
+    """
+    assert s.is_primary_scheduler
+    secondary_scheduler = None
+    if s.secondary_scheduler_name:
+        secondary_scheduler = ilib.await_node_hostname(
+            s.config, s.secondary_scheduler_name
+        )
+    state_save_location = "/var/spool/slurmd"
```
aditigaur4, line 307: should we be calling scheduler's state save location as "slurmd"? that sounds a bit confusing, could it be "slurmctld"?

slurm/install/installlib.py
```diff
@@ -1,16 +1,25 @@
...
 import os
 import pwd
 import re
 import shutil
+from ssl import SSLContext
+import ssl
 import subprocess
 import tempfile
-import time
+from time import sleep as _sleep
```
aditigaur4, line 14: Just trying to understand the rationale behind doing this? It seems we are just replacing time.sleep() in code with sleep() but is it serving some other purpose as well?

slurm/install/installlib.py
```diff
@@ -427,3 +493,183 @@ def set_hostname(
...
+    def __str__(self) -> str:
+        return repr(self)
+
+    def __eq__(self, other: object) -> bool:
+        if hasattr(other, "to_dict"):
+            return self.to_dict() == getattr(other, "to_dict")()
+        return False
+
+
+def cluster_status(config: Dict) -> Dict:
```
aditigaur4, line 552: I am a bit concerned about making REST calls here in installlib. Purely because we have been using this lib as a way to install slurm. Is "scalelib" which is already configured to make calls to CC a better place for this? I may be wrong about this, but to me it looks like its a better fit there.

slurm/install/installlib.py
```diff
@@ -427,3 +493,183 @@ def set_hostname(
             "import jetpack.converge as jc; jc._send_installation_status('warning')",
         ],
     )
+
+
+class CCNode:
+    """
+    Simple CycleCloud node representation.
```
aditigaur4, line 500: again maybe a better fit for scalelib? in fact scalelib should already have a node representation.

slurm/install/installlib.py
```diff
@@ -20,6 +29,42 @@ class ConvergeRetry(RuntimeError):
     pass
 
 
+class Clock:
```
aditigaur4, line 32: Same question here, is there some functionality we are trying to create here in this class?

slurm/install/install.py
```diff
@@ -257,11 +287,28 @@ def accounting(s: InstallSettings) -> None:
...
 
 
 def _complete_install_primary(s: InstallSettings) -> None:
+    """
+    Only the primary scheduler should be creating files under /sched.
+    """
+    assert s.is_primary_scheduler
+    secondary_scheduler = None
+    if s.secondary_scheduler_name:
+        secondary_scheduler = ilib.await_node_hostname(
```
aditigaur4, line 304: I see the value in this function-- when we move to wanting slurmdbd in its own node (people may prefer that) this would be very useful. But strictly speaking for backup controller we dont need to do this, backup controller host just needs to be up and running but does not have to be before rest of slurm installation takes place. I just think its going to cause delays in converging the cluster. Maybe this can be done as a final check when everything else has been done, and before we start slurmctld?

aditigaur4 (commented):
Overall some useful stuff here and I think this one I would like to spend some more time testing. But these are my initial thoughts. It would be useful if some of the REST calls can be moved to scalelib, of course i am curious to hear if you have a better reason for doing it here instead.

slurm/install/install.py
ryanhamel: Good call, slurmctld makes more sense, but we shipped with this in 3.0.3 and in 2.x. Maybe we should create a symlink to /var/spool/slurmd + do a check on their slurm config to tell them this is deprecated?

slurm/install/installlib.py
ryanhamel: It is used so we can inject a mock clock for testing. See installlib_test.py

slurm/install/installlib.py
ryanhamel: As we discussed offline, I do understand that concern but scalelib requires quite a few dependencies and was not itself meant to be used just as an installer. Requiring so many packages to be installed on execute nodes in a CX environment just to install slurm (but not azslurm which uses scalelib) is just not a tradeoff we can make at this time.

slurm/install/install.py
ryanhamel: Yes, we definitely need to handle a true topology to be created but for this version we are simply supporting HA alone. People who require HA + an external slurmdbd will have to manage that after converge for this release.

slurm/install/installlib.py
ryanhamel: See related comment

slurm/install/install.py
aditigaur4: I am not sure about that, i mean i think just fixing it should be enough..The reason is var/spool/slurmd is a valid location that is used for slurmd-- granted slurmd is not running alongside slurmctld so we wont run into issues. I think its not necssary to symlink though because this just reads like a typo or a mistake in porting over configs.

slurm/install/installlib.py
aditigaur4: I dont think the intent here is to use scalelib as an installer. there are a couple of issues i am trying to point out here:

First, it is that installlib should just be concerned with doing installation of rpms, general configuration, and enabling/disabling services. Based on the change we just merged, PR #165, Its not even concerned with starting stopping services. So currently the code in installlib is indempotent, and this change breaks that.

my concern is more with blocking on different nodes in the topology to come up, and second is the obvious code duplication we are having to do here because we cannot include scalelib.

everything we are doing in installlib, can perfectly happen without waiting on the ha node to come up. And based on that we dont really need to have this code at all.

What we can do instead is-- in start_services.sh, just ping the HA node (based on the slurm config) to see if its pingable. and if it is then start slurmctld. Then any failure automatically corresponds to just failure of starting stopping services. Failure of HA node to come up is no different than failing to start slurmdbd (to give an example), and is a perfectly valid reason to fail the script. 

that change would make this PR much more cleaner and self contained, and avoid code duplication.

aditigaur4: I forgot to mention, I fully agree with your assessment that scalelib is not the right place, but what i am suggesting is that this code need not exist at all.

aditigaur4:
Looks good!

PR #169, opened 2023-09-14
Handle cluster names with spaces

ryanhamel:
Fixes issues with spaces in cluster names, as well as making sure the scheduler hostname is lowercased.


aditigaur4:
I don't like the fact that escape function has to be written twice for this.

aditigaur4 (commented):
LGTM

PR #192, opened 2023-11-28
BugFix: Create gres.conf file and populate it.

aditigaur4:
Gres.conf was not getting created since the file did not exist, the linking was not working. Gres.conf also needs to be written on cluster initialization, along side azure.conf.

Also adds a "--no-restart" flag to azslurm scale command. As azslurm scale command is used to write both the partitions file and gres.conf file, but it cannot restart slurmctld during installation because of delayed services block.

Note: This does not change behaviour of azslurm scale command.


ryanhamel (commented):
Nice catch, works as promised.

PR #194, opened 2023-12-12
Fixing incorrect if statements

edwardsp:
The single brackets just meant that it is writing a file called "19" to the current working directory.  It also always evaluates to true unless UBUNTU_VERSION is empty.


aditigaur4:
Good catch! thanks for PR!

Issue #203, opened 2024-01-29
Ubuntu 22+: Hostnames do not register in DNS


ryanhamel:
Due to an issue with the underlying DNS registration scheme that is used across Azure, our Slurm scripts must now use a mitigation that involves restarting systemd-networkd when changing the hostname of VMs deployed in a VMSS. This mitigation should be disabled by adding the following to your Configuration section.

      [[[configuration]]]
      slurm.ubuntu22_waagent_fix = false

PR #205, opened 2024-01-29
Ubuntu22: Mitigation for waagent's failure register DNS

ryanhamel:
Ubuntu 22 or greater and DNS hostname resolution
Due to an issue with the underlying DNS registration scheme that is used across Azure, our Slurm scripts use a mitigation that involves restarting `systemd-networkd` when changing the hostname of VMs deployed in a VMSS. This mitigation can be disabled by adding the following to your `Configuration` section.

      [[[configuration]]]
      slurm.ubuntu22_waagent_fix = false

In future releases, this mitigation will be disabled by default when the issue is resolved in `waagent`.
Fixes: #203

### DIFF
--- a/slurm/install/install.py
+++ b/slurm/install/install.py
@@ -92,6 +92,8 @@ class InstallSettings:
         self.secondary_scheduler_name = config["slurm"].get("secondary_scheduler_name")
         self.is_primary_scheduler = config["slurm"].get("is_primary_scheduler", self.mode == "scheduler")
         self.config_dir = f"/sched/{self.slurm_cluster_name}"
+        # Leave the ability to disable this.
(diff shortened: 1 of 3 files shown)


slurm/install/install.py
```diff
@@ -507,6 +509,28 @@ def set_hostname(s: InstallSettings) -> None:
     ilib.set_hostname(
         new_hostname, s.platform_family, s.ensure_waagent_monitor_hostname
     )
+    if _is_at_least_ubuntu22() and s.ubuntu22_waagent_fix:
```
aditigaur4, line 512: Is it possible to leave a log message here? About restarting waagent. During Troubleshooting via logs it would be good to know at what point during the install the agent was restarted,

aditigaur4 (approved):

Issue #198, opened 2024-01-18
Allow local override of GPU count


ryanhamel:
To both solve the issue where CycleCloud cannot report the proper number of GPUs, due to incorrect backend data, as well as the use case where a user needs to override the reported number of GPUs, allow the ability to define these using scalelib's built-in default_resources mechanism.
For example, by default we would ship with simply
  "default_resources": [
    {
      "select": {},
      "name": "slurm_gpus",
      "value": "node.gpu_count"
    }
  ],
  
  However, if a user wants to / needs to define Standard_XYZ as having 8 GPUs, they can do that by simply editing /opt/azure/slurm/autoscale.json:
  "default_resources": [
    {
      "select": {"node.vm_size": "Standard_XYZ"},
      "name": "slurm_gpus",
      "value": 8
    },
    {
      "select": {},
      "name": "slurm_gpus",
      "value": "node.gpu_count"
    }
  ],

PR #206, opened 2024-01-29
Allow local override of GPUs

ryanhamel:
Per README.md additions: 
For some regions and VM sizes, some subscriptions may report an incorrect number of GPUs. This value is controlled in `/opt/azure/slurm/autoscale.json`

The default definition looks like the following: 
```json
  "default_resources": [
    {
      "select": {},
      "name": "slurm_gpus",
      "value": "node.gpu_count"
    }
  ],
```
Note that here it is saying "For all VM sizes in all nodearrays, create a resource called `slurm_gpus` with the value of the `gpu_count` CycleCloud is reporting".

A common solution is to add a specific override for that VM size. In this case, `8` GPUs. Note the ordering here is critical - the blank `select` statement will set the default for all possible VM sizes and all other definitions will be ignored. For more information on how scalelib `default_resources` work, the underlying library used in all CycleCloud autoscalers, [see the ScaleLib documentation](https://github.com/Azure/cyclecloud-scalelib?tab=readme-ov-file#resources)

```json
  "default_resources": [
    {
      "select": {"node.vm_size": "Standard_XYZ"},
      "name": "slurm_gpus",
      "value": 8
    },
    {
      "select": {},
      "name": "slurm_gpus",
      "value": "node.gpu_count"
    }
  ],
```

Simply run `azslurm scale` again for the changes to take effect. Note that if you need to iterate on this, you may also run `azslurm partitions`, which will write the partition definition out to stdout. This output will match what is in `/etc/slurm/azure.conf` after `azslurm scale` is run.
Fixes: #198

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,42 @@ And for each nodearray, for example the `htc` array:
 ![Alt](/images/nodearrayedit.png "Edit configuration")
 
 
+### Incorrect number of GPUs
+
+For some regions and VM sizes, some subscriptions may report an incorrect number of GPUs. This value is controlled in `/opt/azurehpc/slurm/autoscale.json`
+
+The default definition looks like the following: 
+```json
+  "default_resources": [
+    {
+      "select": {},
+      "name": "slurm_gpus",
+      "value": "node.gpu_count"
+    }
+  ],
+```
+Note that here it is saying "For all VM sizes in all nodearrays, create a resource called `slurm_gpus` with the value of the `gpu_count` CycleCloud is reporting".
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -199,6 +199,42 @@ And for each nodearray, for example the `htc` array:
 ![Alt](/images/nodearrayedit.png "Edit configuration")
 
 
+### Incorrect number of GPUs
+
+For some regions and VM sizes, some subscriptions may report an incorrect number of GPUs. This value is controlled in `/opt/azure/slurm/autoscale.json`
```
aditigaur4, line 204: the path is `/opt/azurehpc/slurm/autoscale.json`

aditigaur4 (approved):

PR #209, opened 2024-01-31
Refactor the network attached storage configuration

bwatrous:
Refactor the network attached storage configuration blade to allow mounting /sched AND an additional /data FS 
Add support for AMLFS as well as NFS


aditigaur4:
@bwatrous  this PR is regressing the changes, so it needs to be rebased off of master and only then can be merged.

PR #219, opened 2024-03-11
slurm.dampen_memory was ignored

ryanhamel:
Fixes: #30

Related new documentation:

Slurm requires that you define the amount of free memory, after OS/Applications are considered, when reporting memory as a resource. If the reported memory is too low, then Slurm will reject this node. To overcome this, by default we dampen the memory by 5% or 1g, whichever is larger.

To change this dampening, there are two options.
1) You can define `slurm.dampen_memory=X` where X is an integer percentage (5 == 5%)
2) Crate a default_resource definition in the /opt/azurehpc/slurm/autoscale.json file. 
  ```json
    "default_resources": [
    {
      "select": {},
      "name": "slurm_memory",
      "value": "node.memory"
    }
  ],
  ```
  Default resources are a powerful tool that the underlying library ScaleLib provides. [see the ScaleLib documentation](https://github.com/Azure/cyclecloud-scalelib?tab=readme-ov-file#resources)

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,27 @@ A common solution is to add a specific override for that VM size. In this case, 
 
 Simply run `azslurm scale` again for the changes to take effect. Note that if you need to iterate on this, you may also run `azslurm partitions`, which will write the partition definition out to stdout. This output will match what is in `/etc/slurm/azure.conf` after `azslurm scale` is run.
 
+### Dampening Memory
+
+Slurm requires that you define the amount of free memory, after OS/Applications are considered, when reporting memory as a resource. If the reported memory is too low, then Slurm will reject this node. To overcome this, by default we dampen the memory by 5% or 1g, whichever is larger.
+
+To change this dampening, there are two options.
+1) You can define `slurm.dampen_memory=X` where X is an integer percentage (5 == 5%)
+2) Create a default_resource definition in the /opt/azurehpc/slurm/autoscale.json file. 
+  ```json
+    "default_resources": [
+    {
+      "select": {},
+      "name": "slurm_memory",
+      "value": "node.memory"
+    }
+  ],
+  ```
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -234,6 +234,24 @@ A common solution is to add a specific override for that VM size. In this case,
...
+2) Crate a default_resource definition in the /opt/azurehpc/slurm/autoscale.json file. 
+  ```json
+    "default_resources": [
+    {
+      "select": {},
+      "name": "slurm_memory",
+      "value": "node.memory"
+    }
+  ],
+  ```
```
aditigaur4, line 252: This needs to mention here that slurm setting will take precedence. Since its mentioned in other places, we should keep the doc consistent.

slurm/src/slurmcc/partition.py
```diff
@@ -45,11 +46,19 @@ def __init__(
         assert len(buckets) > 0
         self.max_scaleset_size = max_scaleset_size
         self.vcpu_count = buckets[0].vcpu_count
-        slurm_memory = buckets[0].resources.get("slurm_memory")
-        if slurm_memory:
-            self.memory = int(slurm_memory.convert_to("m").value)
+        memory_mb = buckets[0].memory.convert_to("m").value 
+        if dampen_memory is not None:
```
aditigaur4, line 50: I am not seeing dampen_memory being set anywhere. It defined as `None` above in the diff. Should we be comparing `memory_mb`?

aditigaur4: nevermind its set below.

README.md
ryanhamel: Good point, I added a comment about this. Line 256
  **Note:** `slurm.dampen_memory` takes precedence, so the default_resource `slurm_memory` will be ignored if `slurm.dampen_memory` is defined.

aditigaur4 (approved):

PR #227, opened 2024-04-22
Update rhel.sh

igmarti:
adding support for rhel9

I did not fully test this as I do not have rhel9 packages, the repositories and checks will enable redhat 9 and almalinux 9 (should work with rocky).


igmarti:
@microsoft-github-policy-service agree company="Microsoft"

aditigaur4 (commented):
This PR obviously is incomplete so it cannot be merged at the moment. We expect RHEL 9 support to be there when we have a rhel 9 hpc image available in marketplace.

PR #231, opened 2024-05-08
Fix manual suspend and avoid KeepAlive=true resume/resume_fail loop via new azslurm return_to_idle command

ryanhamel:
Fixes issues with manually suspending nodes that do not use nodename==hostname.
Also fixes KeepAlive issues where the nodes would resume/resume_failed in a loop forever.
Adds azslurm return_to_idle command to help with this, rather than the pure shell implementation.

### DIFF
--- a/slurm/src/slurmcc/cli.py
+++ b/slurm/src/slurmcc/cli.py
@@ -323,6 +323,110 @@ class SlurmCLI(CommonCLI):
         node_mgr = self._node_mgr(config, self._driver(config))
         self._shutdown(config, node_list=node_list, node_mgr=node_mgr)
 
+    def return_to_idle_parser(self, parser: ArgumentParser) -> None:
+        parser.set_defaults(read_only=False)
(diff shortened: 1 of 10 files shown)


slurm/src/slurmcc/cli.py
```diff
@@ -293,7 +321,112 @@ def resume_fail(
...
+
+        snodes = slutil.show_nodes()
+        if terminate_zombie_nodes:
+            if "return_to_idle" not in config:
+                config["return_to_idle"] = {}
+            config["return_to_idle"]["terminate-zombie-nodes"] = True
+            
+        SlurmCLI._return_to_idle(config, snodes, ccnodes_by_name, scontrol, node_mgr)
+
+    @staticmethod
```
aditigaur4, line 368: As we discussed we can simplify and refactor  this code a little bit by only returning to idle nodes that don't exist. And taking action (shutdown, or mark down) nodes that need to be shut down.

PR #232, opened 2024-05-08
Add --no-jetpack option to install.sh for non-jetpack installations

ryanhamel:
Fail if no jetpack is found unless the user explicitly sets --no-jetpack

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3,6 +3,7 @@
 # Licensed under the MIT License.
 #
 set -e
+
 if [ $(whoami) != root ]; then
   echo "Please run as root"
   exit 1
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -92,7 +102,10 @@ eval "\$(/opt/azurehpc/slurm/venv/bin/register-python-argcomplete azslurm)" || e
 EOF
 fi
 
-which jetpack || exit 0
```
aditigaur4, line 95: I think earlier we were detecting if jetpack did not exist. Now it seems we are assuming that the script has been passed with the right options. which is great but I think we should still detect if jetpack is present and set the `NO_JETPACK` variable accordingly?

aditigaur4 (approved):

PR #233, opened 2024-05-08
In slurm > 23.11 scontrol commands hang if slurmctld is not running.

aditigaur4:
During installation, since we start services later, scontrol commands appear to be hanging, and especially in HA mode where it attempts to talk to backup slurm controller.

### DIFF
--- a/slurm/src/slurmcc/util.py
+++ b/slurm/src/slurmcc/util.py
@@ -186,29 +186,16 @@ def _node_index_and_pg_as_sort_key(nodename: str) -> Union[str, int]:
 _IS_AUTOSCALE_ENABLED = None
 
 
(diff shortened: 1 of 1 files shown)


slurm/src/slurmcc/util.py
```diff
@@ -196,19 +196,11 @@ def is_autoscale_enabled(subprocess_module: Optional[Any] = None) -> bool:
         subprocess_module = subprocess_module_tmp
```
ryanhamel, line 196: Let's remove the subprocess_module argument and and reference in this method to subprocess. I did a git grep, and we aren't actually ever calling this function with a mock subprocess module or anything.

aditigaur4: Done

ryanhamel (approved):

PR #241, opened 2024-05-21
Add storageloc parameter to prevent multiple slurmdbd's from talking to the database

aditigaur4:
That is a misconfiguration of slurm. Also adds a warning for the users to inform them that they are responsible for purging/archiving the database once cluster is deleted.


aditigaur4:
Screenshot 2024-05-20 181027

ryanhamel (approved):

Issue #126, opened 2023-06-08
RHEL setup does not work due to powertools


jpROC1:
RHEL node setup does not work as power tools does not exist on RHEL

jgregorcmh:
@jpROC1 this was also an issue in the older 2.X templates and exists in 3.X.  From our clusters, also running RHEL, perl-Switch package is natively available in RHEL8 from "codeready-builder-for-rhel-8-x86_64-rpms" repo and in RHEL7 from "rhel7-server-optional-rpms".  Those repos are already enabled in the images we are using but can be added with subscription manager.  I've had to add several switches to our custom code to further handle RHEL versus Centos/Alama/Rocky differences with EPEL.

PR #259, opened 2024-06-14
Fix powertools installation for RHEL

diodfr:
This PR is intended to fix issue #126 .
The powertools repository is not compatible nor required for RHEL only for CentOS


aditigaur4:
I ll test this locally, and we'll bring it in before the next release

ryanhamel (approved):

PR #266, opened 2024-07-09
Remove unused C job_submit_plugin to speed up install and resolve codeQL complaints

bwatrous:


.github/workflows/codeql.yml
```diff
@@ -15,7 +15,7 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ "master" ]
+    branches: [ "master", "bewatrou/remove_c_job_submit_plugin" ]
```
ryanhamel, line 18: Was this intended to be left in?

bwatrous: Good catch!

PR #268, opened 2024-07-09
remove requests from setup.py/dev-requirements

ryanhamel:


bwatrous (commented):
For this one... Should we keep the explicit dependency and just bump the version?
Seems like this project directly depends on requests:

slurm/src/slurmcc/__init__.py:import requests
slurm/src/slurmcc/__init__.py:            [RuntimeError, ValueError, requests.exceptions.ConnectionError]

ryanhamel:
That use if requests is only for live testing with a chaos mode which would have scalelib installed. I now handle missing requests library if that module is imported without requests.

PR #283, opened 2024-10-07
Updating cloud bursting scripts and repo using CycleCloud 8.4 and Slurm 23.11.9-1

vinil-v:
Updating cloud bursting setup scripts and repo for creating CycleCloud Slurm Hybrid HPC setup.
- cyclecloud 8.4
- cylecloud-slurm 3.0.9
- slurm 23.11.9-1


vinil-v:
@microsoft-github-policy-service agree company="Microsoft

vinil-v:
@aditigaur4  - could you please review the PR?

xpillons:
@vinil-v can you add an architecture diagram as well as defining which communication port should be open between the on-prem environment and the cloud environment ? I guess that the home directories are still on-prem ? How does users's identity is preserved between both environments ?

vinil-v:
> @vinil-v can you add an architecture diagram as well as defining which communication port should be open between the on-prem environment and the cloud environment ? I guess that the home directories are still on-prem ? How does users's identity is preserved between both environments ?

@xpillons  - updated as per your recommendation.

aditigaur4 (changes requested):
Vinil, Thank you for the PR. The issue is that the PR makes the script hard code the cyclecloud version, the autoscale package version as well as a slurm version. This is not possible for us to support in this manner. Moreover there is a whole new slurm template for this but its unclear why a new slurm template is needed for this. 

We need to do this in a generic way and not locked down to a specific slurm/cyclecloud version, as things will quickly go out of date.

vinil-v:
> Vinil, Thank you for the PR. The issue is that the PR makes the script hard code the cyclecloud version, the autoscale package version as well as a slurm version. This is not possible for us to support in this manner. Moreover there is a whole new slurm template for this, but the better way would be add that in the default templates directory so it gets maintenance updates.
> 
> We need to do this in a generic way and not locked down to a specific slurm/cyclecloud version, as things will quickly go out of date.

@aditigaur4 
Hello 

Aditi

, thank you for reviewing the PR. For this configuration, I utilized the most recent packages from cyclecloud-slurm.

Concerning the version of slurm and the project, version 23.11.9-1 of slurm is the newest in the 3.0.9 project. The main issue is the necessity for an external scheduler to enable this setup. I aligned the slurm version with the cyclecloud-slurm version. However, I've also introduced a variable to manage the project/autoscale and slurm versions, which can be updated with new releases. Could you provide guidance on how to make this setup more adaptable, especially regarding the integration of an external scheduler for headless operations?
About the template - it was derived from cyclecloud-slurm 3.0.9's slurm.txt. To achieve headless functionality, I removed the scheduler components without altering anything else.
To generalize, we should specify the Slurm version alongside the supported project version, possibly creating a compatibility matrix for this configuration. This is under the consideration that we're integrating an external scheduler with our autoscale packages.

PR #292, opened 2024-11-01
Always convert the sort key to an integer when sorting node names

ryanhamel:
Removes str vs int comparison bug by always converting the key to an integer. In this case, simply convert the string as bytes into an integer.

### DIFF
--- a/slurm/src/slurmcc/util.py
+++ b/slurm/src/slurmcc/util.py
@@ -174,7 +174,7 @@ def _node_index_as_sort_key(nodename: str) -> Union[str, int]:
     try:
         return int(nodename.split("-")[-1])
     except Exception:
-        return nodename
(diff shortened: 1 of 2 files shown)


slurm/src/slurmcc/util.py
```diff
@@ -174,7 +174,7 @@ def _node_index_as_sort_key(nodename: str) -> Union[str, int]:
     try:
         return int(nodename.split("-")[-1])
     except Exception:
-        return nodename
+        return int.from_bytes(nodename.encode(), byteorder="little")
 
 
 def _node_index_and_pg_as_sort_key(nodename: str) -> Union[str, int]:
```
aditigaur4, line 180: Does the same change need to be applied in line 189 as well?

aditigaur4:
I need to approve this for now to get the pre-release out but i think this PR might be incomplete.

PR #301, opened 2025-01-13
Make resumeprogram return immediately after resuming node.

aditigaur4:
no-wait option makes the azslurm resume command return immediately and it does not hold up slurmscriptd until the node resumes. Without this option, slurmscriptd process keeps running until the node comes up, which can take 5 minutes or longer. This makes any other commands like "scontrol reconfigure" time out and subsequently all slurm commands timeout.


aditigaur4:
lets hold off on this change for now.

PR #304, opened 2025-01-24
Added GPU nodearray and parameters; updated HTC and HPC to use node counts

abatallas:


templates/slurm.txt
```diff
@@ -195,11 +195,28 @@ Autoscale = $Autoscale
...
+    MachineType = $HTCMachineType
+    ImageName = $HTCImageName
+    MaxCount = $MaxHTCExecuteNodeCount
+
+    Interruptible = $HTCUseLowPrio
+    MaxPrice = $HTCSpotMaxPrice
+    AdditionalClusterInitSpecs = $HTCClusterInitSpecs
+
+        [[[configuration]]]
+        slurm.hpc = true
```
aditigaur4, line 209: This parameter does not need to be true for htc partition.

templates/slurm.txt
```diff
@@ -195,11 +195,28 @@ Autoscale = $Autoscale
         [[[network-interface eth0]]]
         AssociatePublicIpAddress = $ExecuteNodesPublic
 
+    [[nodearray htc]]
```
bwatrous, line 198: Can you re-order this so that git can understand the diffs rather than showing an entirely new nodearray?

templates/slurm.txt
```diff
@@ -277,17 +294,23 @@ Order = 10
         ParameterType = Cloud.MachineType
         DefaultValue = Standard_D8as_v4
 
+        [[[parameter HTCMachineType]]]
```
bwatrous, line 297: Again, try to avoid diffs like this where the only change is moving a block - it makes it hard to differentiate between the meaningful changes and code cleanup.

(When we want to re-order or cleanup the file, that deserves a separate commit.   But this isn't a case that needs re-ordering anyway.)

templates/slurm.txt
```diff
@@ -307,18 +330,26 @@ Order = 10
         Widget.Plugin = pico.form.BooleanCheckBox
         Widget.Label = Start and stop execute instances automatically
 
-        [[[parameter MaxHPCExecuteCoreCount]]]
-        Label = Max HPC Cores
-        Description = The total number of HPC execute cores to start
+        [[[parameter MaxHTCExecuteNodeCount]]]
```
bwatrous, line 333: Sorry to be a broken record...   But same thing here - let's try to keep the ordering the same so the diff makes sense

abatallas: Hey Ben- this is a different parameter. Note cores vs nodes

aditigaur4 (approved):

aditigaur4:
LGTM

PR #306, opened 2025-01-24
Enable nodearray tags for VMSSs

abatallas:


aditigaur4:
This shows up in the cyclecloud UI like this: 
Screenshot 2025-01-24 155847

Is the user supposed to type the tags in between the brackets? can the brackets be removed, so the user can just type in the textbox? Also if there are any azure restrictions on tag names like then should they be mentioned in a comment or something? 

The brackets is making it look a little weird.

templates/slurm.txt
```diff
@@ -296,6 +297,12 @@ Order = 10
         DefaultValue = Standard_F2s_v2
         Config.MultiSelect = true
 
+        [[[parameter NodeTags]]]
+        Label = VM Tags
+        Description = Tags applied to all nodes
+        ParameterType = Record
+        DefaultValue := {}
```
ryanhamel, line 304: The default value should be := []
:= {} is a list
:= [] is a dict

It is unlike most syntaxes

ryanhamel (approved):

PR #314, opened 2025-02-04
Fix slurmd startup for dynamic GPU nodes

aditigaur4:
Dynamic nodes that are precreated need slurmd to start with "gres:gpu:<gpu_count" option, otherwise they get invalid registration in `sinfo` and are unschedulable.

### DIFF
--- a/slurm/install/install.py
+++ b/slurm/install/install.py
@@ -91,10 +91,14 @@ class InstallSettings:
         self.platform_family = platform_family
         self.mode = mode
 
-        self.dynamic_config = config["slurm"].get("dynamic_config")
+        self.dynamic_config = config["slurm"].get("dynamic_config", None)
+        self.dynamic_feature = config["slurm"].get("dynamic_feature", None)
+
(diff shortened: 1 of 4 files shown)


slurm/install/install.py
```diff
@@ -109,19 +109,11 @@ def __init__(self, config: Dict, platform_family: str, mode: str) -> None:
         self.ubuntu22_waagent_fix = config["slurm"].get("ubuntu22_waagent_fix", True)
 
 
-def _inject_vm_size(dynamic_config: str, vm_size: str) -> str:
```
ryanhamel, line 112: We will be breaking backwards compatability with dynamic config by doing this. Can we keep the old _inject_vm_size and rename it _inject_vm_size_deprecated or something?

aditigaur4: Actually in practise this works the same way as before. In fact earlier this wasnt working correctly. In this PR this function actually sets the VM size in the features,

ryanhamel (approved):

PR #318, opened 2025-02-06
Slurm topology configuration tool integration into azslurm cli

azreenz:
Integrating a slurm topology config tool into azslurm cli. Added in test cases and began integrating into cli.py


slurm/src/slurmcc/topology.py
```diff
@@ -0,0 +1,498 @@
...
+        Identifies torsets for hosts based on device GUIDs.
+    group_hosts_by_torset(self) -> dict:
+        Groups hosts by torsets.
+    write_slurm_topology(self, output) -> None:
+        Writes the Slurm topology to a file or prints it.
+    run(self, hosts, partition, output):
+        Executes the entire process of generating the Slurm topology.
+    """
+
+    output_dir: Path
```
aditigaur4, line 84: So these variables that you have defined here are not instance objects. They can be shared across instances of the class. I dont think that is the goal here. So do we need these defined here?

slurm/src/slurmcc/topology.py
```diff
@@ -0,0 +1,498 @@
...
+    guids_file: Path
+    guid_to_host_map: dict = {}
+    hosts_file: Path
+    topo_file: Path
+    slurm_top_file: Path
+    device_guids_per_switch: list = []
+    host_to_torset_map: dict = {}
+    torsets: dict = {}
+    def __init__(self,output):
```
aditigaur4, line 93: You can take more arguments in the init function, since we end up passing more arguments to the run function.

slurm/src/slurmcc/topology.py
```diff
@@ -0,0 +1,498 @@
...
+                  as values.
+        """
+        torsets = {}
+        for host, torset in self.host_to_torset_map.items():
+            if torset not in torsets:
+                torsets[torset] = [host]
+            else:
+                torsets[torset].append(host)
+        return torsets
+    def write_slurm_topology(self,output)-> None:
```
aditigaur4, line 414: You dont need to keep passsing the output variable to the function. Just instantiate it as the member of the class

slurm/src/slurmcc/topology.py
```diff
@@ -0,0 +1,498 @@
...
+        Executes the sequence of steps to generate and write the SLURM topology.
+        Args:
+            hosts (list): List of hostnames to be processed.
+            partition (str): The partition name to which the hosts belong.
+            output (bool): Flag indicating whether to write the output to a file.
+        Returns:
+            None
+        """
+
+        console_handler=logging.StreamHandler()
```
aditigaur4, line 460: Logging related initialization we moved to the logging.conf, so it should not be needed here

azreenz:
@microsoft-github-policy-service agree company="Microsoft"

slurm/src/slurmcc/cli.py
```diff
@@ -184,7 +185,17 @@ def cost(self, config: Dict, start, end, out, fmt=None):
         azcost = azurecost(config)
         driver = cost.CostDriver(azcost, config)
         driver.run(start, end, out, fmt)
-
+    
+    def topology_parser(self, parser: ArgumentParser) -> None:
+        group = parser.add_mutually_exclusive_group(required=True)
+        group.add_argument('-n','--nodes', type=str, help="Specify the hostnames for the nodes")
+        group.add_argument('-p,','--partition', type=str, help="Specify the parititon")
```
aditigaur4, line 192: set required to true for partition argument.

slurm/src/slurmcc/cli.py
```diff
@@ -184,7 +185,17 @@ def cost(self, config: Dict, start, end, out, fmt=None):
         azcost = azurecost(config)
         driver = cost.CostDriver(azcost, config)
         driver.run(start, end, out, fmt)
-
+    
+    def topology_parser(self, parser: ArgumentParser) -> None:
+        group = parser.add_mutually_exclusive_group(required=True)
+        group.add_argument('-n','--nodes', type=str, help="Specify the hostnames for the nodes")
```
aditigaur4, line 191: Remove -n nodes argument and remove any validation done for nodes argument

slurm/src/slurmcc/cli.py
```diff
@@ -184,7 +185,17 @@ def cost(self, config: Dict, start, end, out, fmt=None):
         azcost = azurecost(config)
         driver = cost.CostDriver(azcost, config)
         driver.run(start, end, out, fmt)
-
+    
+    def topology_parser(self, parser: ArgumentParser) -> None:
```
aditigaur4, line 189: We will add a new boolean argument called `use_vmss` that is defaulted to `True`. And if `True` (set by default) we will return the output of `_generate_topology` function.

Then add new boolean argument called `use_fabric_manager` defaulted to `False`. But when set to true, then it runs the topology class logic. 

When `use_fabric_manager` is set to True, then `partitions` is a required flag. 

Use_vmss and use_fabric_manager cannot both be true at the same time.

aditigaur4:
Finally, lets rebase down to 1 commit.

slurm/src/slurmcc/util.py
```diff
@@ -4,14 +4,27 @@
...
 import time
 import traceback
 from abc import ABC, abstractmethod
 from typing import Any, Callable, Dict, List, Optional, Union
-
+import sys
 from . import AzureSlurmError, custom_chaos_mode
 
 
+class SrunExitCodeException(Exception):
```
aditigaur4, line 16: What happened here?  I don't think this change belongs in the diff

slurm/src/slurmcc/util.py
```diff
@@ -258,4 +271,19 @@ def is_autoscale_enabled() -> bool:
     if _IS_AUTOSCALE_ENABLED is not None:
         return _IS_AUTOSCALE_ENABLED
     logging.warning("Could not determine if autoscale is enabled. Assuming yes")
-    return True
\ No newline at end of file
+    return True
+
```
aditigaur4, line 275: Replace this function with the `run` function merged just before. After this change, you should have no other changes in util.py

slurm/src/slurmcc/util.py
azreenz: Yeah i'm not sure... maybe something happened in the rebase

slurm/src/slurmcc/topology.py
```diff
@@ -0,0 +1,410 @@
...
+        Determines the appropriate SHARP command based on the operating system.
+
+        Returns:
+            str: The path to the SHARP command for the detected operating system.
+
+        Raises:
+            SystemExit: If the operating system is not supported.
+        """
+        os_id=self.get_os_name()
+        if os_id == "\"ubuntu\"":
```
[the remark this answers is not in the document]
azreenz: i think i fixed this bug can you check on your system if it works

slurm/src/slurmcc/util.py
aditigaur4: Can this be fixed?

slurm/src/slurmcc/topology.py
```diff
@@ -0,0 +1,410 @@
...
+            Warnings for invalid or powered-down nodes.
+            Debug information for the list of hosts and the filtered valid hosts.
+            Error if the number of valid and powered-on hosts is less than 2.
+        """
+        def get_hostlist(cmd) -> list:
+            output=slutil.run(cmd, shell=True)
+            return set(output.stdout.split('\n')[:-1])
+        partition_cmd = f'-p {self.partition} '
+        host_cmd = f'scontrol show hostnames $(sinfo -p {self.partition} -o "%N" -h)'
+        partition_states = "powered_down,powering_up,powering_down,power_down"
```
aditigaur4, line 52: When we first wrote this logic we were doing parallel-ssh so our criteria was just to understand which hosts are pingable in the partition. But now since we are submitting slurm jobs, I suppose we need to change this to "which hosts can run a slurm job". A node at the minimum needs to be idle and powered_up. So this logic of filtering hosts atleast for running slurm jobs needs to be changed.

slurm/src/slurmcc/topology.py
```diff
@@ -0,0 +1,410 @@
...
+            None
+        Raises:
+            SystemExit: If the number of valid and powered-on hosts is less than 2.
+        Logs:
+            Warnings for invalid or powered-down nodes.
+            Debug information for the list of hosts and the filtered valid hosts.
+            Error if the number of valid and powered-on hosts is less than 2.
+        """
+        def get_hostlist(cmd) -> list:
+            output=slutil.run(cmd, shell=True)
```
aditigaur4, line 48: We are not doing any exception handling here. Lets catch atleast timeout expired, and calledprocessError. Logging isnt necessary since the run function takes care of it. sys.exit(1) is fine.

slurm/src/slurmcc/topology.py
```diff
@@ -0,0 +1,410 @@
...
+        Raises:
+            SystemExit: If the sharp_hello command fails, the program exits with the 
+                        corresponding exit code.
+        """
+        cmd = f"{self.sharp_cmd_path}sharp/bin/sharp_hello"
+        try:
+            output = slutil.srun([self.hosts[0]],cmd)
+            log.debug(output.stdout)
+        except slutil.SrunExitCodeException as e:
+            log.error("sharp_hello command failed")
```
aditigaur4, line 150: Here i'd like to see a message: that "Sharp is disabled on this cluster". Because sharp_hello failing is not a bug if its disabled. Right now i am seeing:
2025-02-20 23:04:49,402 ERROR: sharp_hello command failed
2025-02-20 23:04:49,402 ERROR: sharp_coll_init failed: SHARP device error

Which is not helpful to the user as to what they should do.

slurm/src/slurmcc/topology.py
```diff
@@ -0,0 +1,410 @@
...
+        partition_states = "powered_down,powering_up,powering_down,power_down"
+        sinfo_cmd = f'sinfo {partition_cmd}-t {partition_states} -o "%N" -h'
+        down_cmd = f'scontrol show hostnames $({sinfo_cmd})'
+        hosts=get_hostlist(host_cmd)
+        down_hosts=get_hostlist(down_cmd)
+        powered_down_hosts = hosts&down_hosts
+        self.hosts = list(hosts-powered_down_hosts)
+        if len(self.hosts)<len(hosts):
+            log.warning(
+                "Some nodes were powered down, powering down, or powering up, "
```
aditigaur4, line 61: Lets re-write this to include all the states that we are excluding.

slurm/src/slurmcc/topology.py
```diff
@@ -0,0 +1,410 @@
...
+                "running on a subset of nodes that are powered on"
+            )
+            if powered_down_hosts:
```
aditigaur4, line 64: Can we log a list of all the hostnames we are excluding? not just the powered_down ones.

README.md
```diff
@@ -323,6 +323,7 @@ This will change the behavior of the `azslurm return_to_idle` command that is, b
       scale                - 
       shell                - Interactive python shell with relevant objects in local scope. Use --script to run python scripts
       suspend              - Equivalent to SuspendProgram, shutsdown nodes
+      topology             - Generates topology plugin configuration
```
aditigaur4, line 326: Can you also remove the generate_topology above? since we removed that CLI

slurm/src/slurmcc/__init__.py
```diff
@@ -10,7 +10,7 @@
 except ImportError:
     # this is only used during live testing with scalelib, so this should never happen
     import logging
-    logging.exception()
+    logging.exception("Failed to import ConnectionError from requests.exceptions")
```
aditigaur4, line 13: Why do we need this change?

slurm/src/slurmcc/cli.py
```diff
@@ -201,11 +228,11 @@ def partitions(self, config: Dict, allow_empty: bool = False) -> None:
             autoscale=is_autoscale_enabled(),
         )
 
-    def generate_topology(self, config: Dict) -> None:
-        """
-        Generates topology plugin configuration
-        """
-        return _generate_topology(self._get_node_manager(config), sys.stdout)
+    # def generate_topology(self, config: Dict) -> None:
```
aditigaur4, line 231: We can just remove this

aditigaur4 (approved):
LGTM

PR #319, opened 2025-02-10
Add an srun interface for submitting parallel commands.

aditigaur4:
commands that need to be run across a range of compute nodes can be submitted through srun. The interface exposed is similar to subprocess but based on srun.

This does use the job scheduler.

### DIFF
--- a/slurm/src/slurmcc/util.py
+++ b/slurm/src/slurmcc/util.py
@@ -1,7 +1,9 @@
 import logging
 from abc import ABC, abstractmethod
+import os
 import random
 import subprocess as subprocesslib
+import tempfile
 import time
 import traceback
 from abc import ABC, abstractmethod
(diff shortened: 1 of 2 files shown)


slurm/src/slurmcc/util.py
```diff
@@ -24,6 +41,24 @@ def scontrol(self, args: List[str], retry: bool = True) -> str:
             return retry_subprocess(lambda: check_output(full_args)).strip()
         return check_output(full_args).strip()
 
+    def srun(self, hostlist: List[str], user_command: str, timeout: int) -> SrunOutput:
+        with tempfile.NamedTemporaryFile(delete=False) as temp_file:
+            temp_file_path = temp_file.name
+
```
ryanhamel, line 47: Why not just put the block under the with, and get rid of the finally: os.remove?

aditigaur4: good catch, fixed!

ryanhamel (approved):

PR #330, opened 2025-02-26
Add bug fixes to slurm topology cli tool 

azreenz:
fix guids parsing
add in SHARP_SMX_UCX_INTERFACE env variable to sharp command
integrate partition option in srun commands
edit topology test nodes_guids input to mimic how cluster outputs data
cleaning up comments
add more logging


aditigaur4:
1. Add a comment in the topology output that adds the number of nodes in above switch, `#Number of Nodes in sw00`
2. Send a SIGTERM to the srun process when it times out.

aditigaur4 (approved):

PR #331, opened 2025-02-27
Add support for Availability Zones

wolfgang-desalvador:
Add support for specifications of Availability Zones on Nodearrays

### DIFF
--- a/templates/slurm.txt
+++ b/templates/slurm.txt
@@ -203,10 +203,11 @@ Autoscale = $Autoscale
     MachineType = $HPCMachineType
     ImageName = $HPCImageName
     MaxCount = $MaxHPCExecuteNodeCount
+    Zone = ${ifThenElse(DefineNodesAvailabilityZone, HPCAvailabilityZone, {})}
     Azure.MaxScalesetSize = $HPCMaxScalesetSize
     AdditionalClusterInitSpecs = $HPCClusterInitSpecs
     EnableNodeHealthChecks = $EnableNodeHealthChecks
-
+    
 
         [[[configuration]]]
         slurm.default_partition = true
@@ -218,7 +219,7 @@ Autoscale = $Autoscale
     MachineType = $HTCMachineType
     ImageName = $HTCImageName
     MaxCount = $MaxHTCExecuteNodeCount
-
+    Zone = ${ifThenElse(DefineNodesAvailabilityZone, HTCAvailabilityZone, {})}
     Interruptible = $HTCUseLowPrio
     MaxPrice = $HTCSpotMaxPrice
     AdditionalClusterInitSpecs = $HTCClusterInitSpecs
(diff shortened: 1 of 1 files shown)


templates/slurm.txt
```diff
@@ -203,10 +203,11 @@ Autoscale = $Autoscale
     MachineType = $HPCMachineType
     ImageName = $HPCImageName
     MaxCount = $MaxHPCExecuteNodeCount
+    Zone = ${ifThenElse(DefineNodesAvailabilityZone, HPCAvailabilityZone, {})}
```
bwatrous, line 206: I'd prefer this to just set Zone to undefined if user did not select a zone.
So just "Zone = $HPCAvailabilityZone" (and we should make sure that the Zones default to undefined)

templates/slurm.txt
```diff
@@ -418,6 +420,30 @@ Order = 10
         Conditions.Excluded := DynamicUseLowPrio isnt true
         Config.MinValue = -1
 
+        [[[parameter DefineNodesAvailabilityZone]]]
+        Label = Specify Availability Zones
+        DefaultValue = false
+        Widget.Plugin = pico.form.BooleanCheckBox
+        Widget.Label = Specify availability zones for nodes
+        
+        [[[parameter HPCAvailabilityZone]]]
```
bwatrous, line 429: I'd set DefaultValue=undefined for these just to make it clear that they're not applied.

bwatrous (commented):
Have you tested this with more than one AZ allowed for a nodearray?   (It _should_ work - i've just never tried it)

templates/slurm.txt
wolfgang-desalvador: Added "undefined" as fallback value

templates/slurm.txt
wolfgang-desalvador: For StringList, undefined or {} as default value create artifacts in the UI. Kept undefined in the fallback above

wolfgang-desalvador:
> Have you tested this with more than one AZ allowed for a nodearray? (It _should_ work - i've just never tried it)
Yes, I have tested it and the multiple AZ are applied correctly to the Virtual Machine ScaleSet.

image

The only thing we need to validate -> for HPC and GPU partition, since they are configured for a single placement group, selecting multiple Availability Zone leads (correctly) to an error in VMSS allocation -> 

`Virtual Machine Scale Set deployed to multiple Availability Zones with 'platformFaultDomainCount' set to 1 and Single Placement Group set to true is not supported.`

Do you think we should:
* Add a warning message on the top?
* Switch the HPC/GPU Availability zones to be a dropdown list?

ryanhamel (dismissed):

abatallas:
@azreen We'll need your eyes on this one too 👍

ryanhamel (approved):

PR #332, opened 2025-03-03
Fix srun logging to output all stderr

azreenz:
edit srun interface to include stderr of srun process and srun command. Fix logging in topology.py to to log every line of stderr output from SrunExitCodeException.

### DIFF
--- a/slurm/src/slurmcc/topology.py
+++ b/slurm/src/slurmcc/topology.py
@@ -110,6 +110,8 @@ class Topology:
             stdout=output.stdout
         except slutil.SrunExitCodeException as e:
             log.error("Error running get_os_id command on host %s",self.hosts[0])
+            if e.stderr_content.strip("\n"):
+                log.error(e.stderr_content)
             log.error(e.stderr)
             sys.exit(e.returncode)
         except subprocesslib.TimeoutExpired:
@@ -166,6 +168,8 @@ class Topology:
             log.debug(output.stdout)
         except slutil.SrunExitCodeException as e:
             log.error("SHARP is disabled on cluster")
+            if e.stderr_content.strip("\n"):
+                log.error(e.stderr_content)
             log.error(e.stderr)
             sys.exit(e.returncode)
         except subprocesslib.TimeoutExpired:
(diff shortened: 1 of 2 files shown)


slurm/src/slurmcc/topology.py
```diff
@@ -225,7 +228,8 @@ def retrieve_guids(self) -> None:
             output = slutil.srun(self.hosts, cmd, shell=True, partition=self.partition)
         except slutil.SrunExitCodeException as e:
             log.error("Error running retrieve_guids command on hosts")
-            log.error(e.stderr)
+            for line in e.stderr.split('\n')[:-1]:
+                log.error(line)
```
aditigaur4, line 232: If our exception just returns stderr_content then we can just log it here without doing additional processing.

slurm/src/slurmcc/util.py
```diff
@@ -61,7 +61,7 @@ def srun(self, hostlist: List[str], user_command: str, timeout: int, shell: bool
                 logging.error(f"Command: {srun_command} failed with return code {e.returncode}")
                 with open(temp_file_path, 'r') as f:
                     stderr_content = f.read()
-                raise SrunExitCodeException(returncode=e.returncode,stdout=e.stdout, stderr=stderr_content)
+                raise SrunExitCodeException(returncode=e.returncode,stdout=e.stdout, stderr=f"{e.stderr}\n{stderr_content}")
```
aditigaur4, line 64: Can we just add another field to srunexitcodeexception called `stderr_content`?

slurm/src/slurmcc/topology.py
```diff
@@ -110,6 +110,8 @@ def get_os_name(self):
             stdout=output.stdout
         except slutil.SrunExitCodeException as e:
             log.error("Error running get_os_id command on host %s",self.hosts[0])
+            if e.stderr_content.strip("\n"):
```
aditigaur4, line 113: If this line needs to be repeated everytime, then it needs to be a part of the srun interface.

azreenz: do you want me to put the if statement in the srun interface or just the .strip("\n") part

PR #333, opened 2025-03-04
Fix do-install flag.

aditigaur4:
If packages are installed, slurm configuration still needs to be done.


aditigaur4:
I think i have some ideas on some more work that can make this bit more refined...stay tuned, marking it draft for now

PR #341, opened 2025-03-19
Move files according to the package they are a part of.

aditigaur4:
repo structure needed to be improved, files that are a part of one package need to be in that package.


.gitignore
```diff
@@ -341,8 +341,8 @@ blobs
 **/dist
 **/*egg-info
 **/creds.json
-slurm/install/*.zip
+azure-slurm-install/dist
 slurm/install/AzureCA.pem
```
azreenz, line 345: did you want to change line 345 to azure-slurm-install too?

aditigaur4: Fixed

azreenz (approved):

PR #350, opened 2025-04-08
Add block topology configuration tool to azslurm cli

azreenz:
Integrate block topology tool into topology config generator 
map hostnames to clusterUUID to identifiy racks for each hostname
create block for each rack
Modify test cases to validate block topology tool


aditigaur4:
This needs some more work.

azure-slurm/slurmcc/cli.py
```diff
@@ -192,7 +192,8 @@ def topology_parser(self, parser: ArgumentParser) -> None:
         parser.add_argument('-o', '--output', type=str, help="Specify slurm topology file output")
         group.add_argument('-v', '--use_vmss', action='store_true', default=True, help='Use VMSS (default: True)')
         group.add_argument('-f', '--use_fabric_manager', action='store_true', default=False, help='Use Fabric Manager (default: False)')
-    def topology(self, config: Dict, partition, output, use_vmss, use_fabric_manager):
+        group.add_argument('-b', '--use_block_topology', action='store_true', default=False, help='Use Block Topology (default: False)')
```
aditigaur4, line 195: The argument here is wrong.

Block topology is just a slurm configuration file. You can use any backend data to generate any slurm configuration file. So you are not "using block topology". you are just outputting block topology here. Technically you could have outputted a block topology file from the above options as well.

what that means is that which slurm configuration file you generate (output) is different from which input you use to generate that file. Previous options were VMSS, fabric manager or in this case clique ID.

aditigaur4: Basically what you need is another option that defines what topology we need to generate. Do we need block topology or do we need tree topology. The current option just needs to be renamed to "use-nvlink-domain". And then do not just assume the output has to be block topology.

azure-slurm/slurmcc/cli.py
```diff
@@ -188,28 +188,57 @@ def cost(self, config: Dict, start, end, out, fmt=None):
...
+        topology_group.add_argument('-t', '--tree', action='store_true', default=False, help='Generate Tree Topology output to use Tree topology plugin with VMSS aligned "racks"  (default: False)')
+
+    def topology(self, config: Dict, partition, output, use_vmss, use_fabric_manager, use_nvlink_domain, tree, block) -> None:
         """
         Generates Topology Plugin Configuration
         """
         if use_fabric_manager:
             if not partition:
                 raise ValueError("--partition is required when using --use_fabric_manager")
+            if not tree and not block:
```
aditigaur4, line 207: Do you really need 3 if else loops to do these checks? You only need to check if topology is block to raise a value error and default it to tree.

azure-slurm/slurmcc/cli.py
```diff
@@ -188,28 +188,57 @@ def cost(self, config: Dict, start, end, out, fmt=None):
...
+                topo_type = "tree"
+            elif block:
+                raise ValueError("--block is not supported with --use_fabric_manager")
+            config_dir = config.get("config_dir")
+            topo = topology.Topology(partition,output,"fabric",topo_type,config_dir)
+            topo.run()
+        elif use_nvlink_domain:
+            if not partition:
+                raise ValueError("--partition is required when using --use_nvlink_domain")
+            if not tree and not block:
```
aditigaur4, line 219: again 3 if-else loops is completely unnecessary here

azure-slurm/slurmcc/cli.py
```diff
@@ -188,28 +188,57 @@ def cost(self, config: Dict, start, end, out, fmt=None):
...
+            elif tree:
+                raise ValueError("--tree is not supported with --use_nvlink_domain")
+            elif block:
+                topo_type = "block"
             config_dir = config.get("config_dir")
-            topo = topology.Topology(partition,output,config_dir)
+            topo = topology.Topology(partition,output,"nvlink",topo_type,config_dir)
             topo.run()
         elif use_vmss:
+            if not tree and not block:
```
aditigaur4, line 229: same this needlessly complicated logic that can be accomplished in 1 statement

aditigaur4: this is what you are doing here: `topo_type = "block" if block else "tree"`

azure-slurm/slurmcc/cli.py
```diff
@@ -188,28 +188,57 @@ def cost(self, config: Dict, start, end, out, fmt=None):
...
-    def topology(self, config: Dict, partition, output, use_vmss, use_fabric_manager):
+        group.add_argument('-v', '--use_vmss', action='store_true', default=True, help='Use VMSS to map Tree or Block topology along VMSS boundaries without special network consideration (default: True)')
+        group.add_argument('-f', '--use_fabric_manager', action='store_true', default=False, help='Use Fabric Manager to map Tree topology (Block topology not allowed) according to SHARP network topology tool(default: False)')
+        group.add_argument('-n', '--use_nvlink_domain', action='store_true', default=False, help='Use NVlink domain to map Block topology (Tree topology not allowed) according to NVLink Domain and Partition for multi-node NVLink (default: False)')
+        topology_group.add_argument('-b', '--block', action='store_true', default=False, help='Generate Block Topology output to use Block topology plugin with NVLink or VMSS aligned blocks (default: False)')
+        topology_group.add_argument('-t', '--tree', action='store_true', default=False, help='Generate Tree Topology output to use Tree topology plugin with VMSS aligned "racks"  (default: False)')
```
aditigaur4, line 198: Why does the help for tree topology needs to indicate anything about VMSS?

azure-slurm/slurmcc/cli.py
```diff
@@ -188,28 +188,57 @@ def cost(self, config: Dict, start, end, out, fmt=None):
...
-        group.add_argument('-f', '--use_fabric_manager', action='store_true', default=False, help='Use Fabric Manager (default: False)')
-    def topology(self, config: Dict, partition, output, use_vmss, use_fabric_manager):
+        group.add_argument('-v', '--use_vmss', action='store_true', default=True, help='Use VMSS to map Tree or Block topology along VMSS boundaries without special network consideration (default: True)')
+        group.add_argument('-f', '--use_fabric_manager', action='store_true', default=False, help='Use Fabric Manager to map Tree topology (Block topology not allowed) according to SHARP network topology tool(default: False)')
+        group.add_argument('-n', '--use_nvlink_domain', action='store_true', default=False, help='Use NVlink domain to map Block topology (Tree topology not allowed) according to NVLink Domain and Partition for multi-node NVLink (default: False)')
+        topology_group.add_argument('-b', '--block', action='store_true', default=False, help='Generate Block Topology output to use Block topology plugin with NVLink or VMSS aligned blocks (default: False)')
```
aditigaur4, line 197: Same. Why does it need to indicate anything about nvlink or VMSS?
I can use block topology on regular F2 VM's.

azure-slurm/slurmcc/cli.py
```diff
@@ -864,7 +893,10 @@ def _generate_topology(node_mgr: NodeManager, writer: TextIO) -> None:
             continue
         nodes = sorted(nodes, key=slutil.get_sort_key_func(bool(pg)))
         slurm_node_expr = ",".join(nodes)  # slutil.to_hostlist(",".join(nodes))
-        writer.write("SwitchName={} Nodes={}\n".format(pg or "htc", slurm_node_expr))
+        if topo_type == "tree":
+            writer.write("SwitchName={} Nodes={}\n".format(pg or "htc", slurm_node_expr))
+        elif topo_type == "block":
```
aditigaur4, line 898: Will this ever generate a valid block_topology file? i dont think it will because it needs a block size

azure-slurm/slurmcc/topology.py
```diff
@@ -406,11 +488,42 @@ def write_slurm_topology(self)-> None:
...
+        Executes the NVLink topology discovery process for the cluster.
+
+        This method performs the following steps:
+            1. Retrieves the list of hostnames in the cluster.
+            2. Obtains rack IDs for each host.
+            3. Groups hosts by their respective racks and stores the result.
+
+        Logging is performed at each major step for debugging purposes.
+        """
+        log.debug("Retrieving hostnames")
```
aditigaur4, line 503: This much logging after every statement is adding no real value. If you just want to add a logging for every function that ever gets called, you can add a decorator.

azure-slurm/slurmcc/topology.py
```diff
@@ -433,9 +546,77 @@ def run(self):
         log.debug("Identified torsets for hosts")
         self.torsets = self.group_hosts_by_torset()
         log.debug("Finished grouping hosts by torsets")
-        self.write_slurm_topology()
-        if self.topo_file:
-            log.info("Finished writing slurm topology from torsets to %s",
+    
+    def run_tree(self):
```
aditigaur4, line 550: When will this function get called?

azure-slurm/slurmcc/topology.py
```diff
@@ -433,9 +546,77 @@ def run(self):
...
+            if self.topo_type == "block":
+                self.run_block()
+            elif self.topo_type == "tree":
+                self.run_tree()
+            else:
+                log.error("Invalid topology type specified. Exiting.")
+                sys.exit(1)
+        elif self.topo_input=="fabric":
+            self.run_fabric()
+            if self.topo_type == "block":
```
aditigaur4, line 612: again you need better code organization here. A common function that generates topology based on an argument whether it needs to generate block or tree

azure-slurm/slurmcc/topology.py
```diff
@@ -433,9 +546,77 @@ def run(self):
...
+        Depending on the value of `self.topo_input`, this method selects the appropriate
+        topology setup routine (`run_nvlink` or `run_fabric`). After setting up the base
+        topology, it further refines the configuration based on `self.topo_type`, calling
+        either `run_block` or `run_tree`. If an invalid topology type or input is provided,
+        an error is logged and the program exits.
+
+        Raises:
+            SystemExit: If an invalid topology input or type is specified.
+        """
+        if self.topo_input=="nvlink":
```
aditigaur4, line 601: same comment as below. You are repeating code that does not need to be repeated.

aditigaur4: Why are these if else loops being repeated over and over? the same check was being done in the CLI. but even if you want to do the check again what is this check exactly checking?

aditigaur4: You are basically checking if someone passed in some topology like "xyz" because your cli function passes in strings for topology. Why isnt it an enumerated type? why arent you using enums such that topology type can only ever be "BLOCK" or "TREE"

azure-slurm/slurmcc/cli.py
bwatrous: Good question, but currently -b and -t ONLY apply when --use_vmss is ALSO selected.
The args are pretty confusing (which is why i asked Azreen to improve the help).
It feels like an alternative would be to remove the --use_fabric_manager and --use_nvlink_domain options and default to the appropriate one of those _unless_ --use_vmss was specified or if it would conflict with the selected -b or -t option.   (Ie - it feels like we could probably default to the "right" selection in general, and allow users to over-ride block vs tree and sku-native vs VMSS aligned)

bwatrous: another alternative would be to switch to an Enum style option if we think being specific about the hardware topology is important :
--group_by=[ vmss | fabric_manager | nvlink_domain | etc. ]

azure-slurm/slurmcc/cli.py
bwatrous: Could potentially use the value of Azure.MaxScaleSetSize as the block size - ie - scalelib knows the maximum number of nodes per block with VMSS aligned blocks.

azure-slurm/slurmcc/cli.py
aditigaur4: I definitely agree that help option for use-vmss/use-fabric etc are confusing and those be elaborated as to what topology types they support. But that is not a restriction on what tree topology means here. Tree/block topology definition inherently doesn't care where the data comes from (fabric_manager, vmss or sharp etc).

azure-slurm/slurmcc/topology.py
```diff
@@ -370,7 +446,33 @@ def group_hosts_by_torset(self) -> dict:
...
+        Writes the block topology configuration to a file or prints it to the console.
+
+        This method generates the block topology configuration based on the `torsets` attribute
+        and either writes it to a file specified by `self.topo_file` or prints it to the console,
+        depending on the value of the `output` parameter.
+
+        Returns:
+            None
+        """
+        if self.slurm_top_file:
```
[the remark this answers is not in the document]
bwatrous: but I think I'd rather just dump to string - it will be more easily testable than writing to file and then having to read the file.

azure-slurm/slurmcc/topology.py
```diff
@@ -6,29 +6,53 @@
...
+    def __str__(self):
+        return self.value
 class Topology:
     """
     A class to represent and manage the topology of a Slurm cluster.
     Attributes:
     """
 
-    def __init__(self,partition, output,directory):
+    def __init__(self,partition,output,topo_input,topo_type,directory):
```
bwatrous, line 34: just for readability use more whitespace (space after commas, and blank line between method defs)

azure-slurm/test/slurmcc_test/topology_test.py
```diff
@@ -179,7 +185,7 @@ def test_get_os_name():
     5. Assert that the result is 'ubuntu'.
     """
     slutil.srun=run_parallel_cmd
-    test_obj   = Topology("hpc",None,TESTDIR)
+    test_obj   = Topology("hpc",None,"fabric","tree",TESTDIR)
     test_obj.hosts=['node1']
     result=test_obj.get_os_name()
```
bwatrous, line 190: It makes me sad that CC doesn't give the base os name in the node status output.  
But it's still not worth doing an srun just for get_os_name to pick a path for following srun commands.
Following commands should simply srun a script that checks the os on the node that it runs on.   This is also safer than just checking the first node in a partition.

bwatrous: but i bet we can do better - there's probably a module that can be loaded to get the proper sharp environment on the target node

azure-slurm/test/slurmcc_test/topology_test.py
```diff
@@ -32,7 +32,7 @@ def __init__(self,stdout,stderr,exit_code):
         self.stdout=stdout
         self.stderr=stderr
         self.returncode=exit_code
-def run_parallel_cmd(hosts,cmd,shell=False, partition=None):
+def run_parallel_cmd(hosts,cmd,shell=False, partition=None,gpus=None):
```
bwatrous, line 35: This is a cool idea for how to return expected results from the srun commands :)

But it's very, very fragile and hard to maintain.
Instead, consider extracting the "cmd + srun"  lines from the Topology class methods into private methods.  If you did that, then you just monkey patch the private method in your test and your tests don't need to know the specific bash command being run.

azure-slurm/test/slurmcc_test/topology_test.py
bwatrous: I see places later in the tool where you use:
`                f"{env['SHARP_CMD']}sharp/bin/sharp_cmd topology "
`
That could work to simplify the cmd above too

azure-slurm/test/slurmcc_test/topology_test.py
aditigaur4: The reason we did it this was because we wanted to be able swap out srun class with any other interface we may use down the road like parallel-ssh or clush. So we did not want to enclose too much custom logic that is slurm specific into the topology class. Topology class basically just runs a few commands in bash through the run_parallel_cmd interface which can point to anything else. I personally am not a fan of the srun interface but it was the best choice without having to set up root ssh keys.

azure-slurm/test/slurmcc_test/topology_test.py
aditigaur4: The only disadvantage to run scripts is it can be hard to capture errors like this in the srun interface. We could use sbatch but it does not provide the most friendly interactive interface. We are basically using srun like how you would use parallel-ssh.

azure-slurm/test/slurmcc_test/topology_test.py
```diff
@@ -291,7 +297,7 @@ def test_generate_topo_file():
                         expected content.
     """
     slutil.srun=run_parallel_cmd
-    test_obj   = Topology("hpc",None,TESTDIR)
+    test_obj   = Topology("hpc",None,"fabric","tree",TESTDIR)
     test_obj.hosts=['node1']
     test_obj.generate_topo_file()
```
bwatrous, line 302: How does this method work in practice?   it looks like `self.topo_file` will be generated on the first node in the partition, but the output is just logged at the scheduler.  

This is one reason that I'd prefer that tests do not rely on reading generated files for most tests.

azure-slurm/test/slurmcc_test/topology_test.py
```diff
@@ -300,39 +306,66 @@ def test_generate_topo_file():
...
-    test_obj   = Topology("hpc",output,TESTDIR)
+    test_obj   = Topology("hpc",output,"fabric","tree",TESTDIR)
     test_obj.get_hostnames()
     test_obj.retrieve_guids()
     test_obj.topo_file = 'test/slurmcc_test/topology_test_input/topology.txt'
     test_obj.device_guids_per_switch =  test_obj.group_guids_per_switch()
     test_obj.host_to_torset_map = test_obj.identify_torsets()
     test_obj.torsets = test_obj.group_hosts_by_torset()
-    test_obj.write_slurm_topology()
+    test_obj.write_tree_topology()
```
bwatrous, line 334: Another reason to split the `write_tree|block_topology()` methods into a 
`topo_str = generate_tree|block_topology()
do_something_with_topology(topo_str)
`
then you could test the generate method with many inputs trivially by just asserting against the "topo_str" output without a ton of golden input/output files.

azure-slurm/test/slurmcc_test/topology_test.py
azreenz: I can look into this further in a separate PR if needed but i dont think resolving this is required for this PR since its separate from block topology

azure-slurm/test/slurmcc_test/topology_test.py
azreenz: For the block topology tests, i added the suggestion of monkey patching the private methods in the test. I had a question about how to take in test input, some of my test input are long ie. 72 lines (to mimic 4 racks of 18 gb200 nodes) do you want me to input this directly in topology_test.py to not rely on reading files or are you only talking about reading output files? Or do you think i should decrease the amount of input i give a test? Another reason as to why i read generated files it to test that output files are being generated correctly since the cli creates a topology.conf.

aditigaur4 (approved):
Looks good

PR #356, opened 2025-05-01
Create prolog.d and epilog.d in config_dir and symlink to /etc/slurm

azreenz:
Create prolog.d and epilog.d in config_dir and symlink to /etc/slurm


aditigaur4 (approved):

aditigaur4 (commented):
One more change needed is to set the Prolog, Epilog in slurm.conf to prolog.d and epilog.d directories. `Prolog="/etc/slurm/prolog.d/*` and similarly for epilog

aditigaur4 (approved):

PR #357, opened 2025-05-02
Integrate IMEX Prolog and Epilog scripts into azure-slurm-install

azreenz:
add imex_prolog.sh and imex_epilog.sh into azure-slurm-install and copy them to /sched/<cluster-name>/prolog.d and /sched/<cluster-name>/epilog.d respectively.

### DIFF
--- a/azure-slurm-install/install.py
+++ b/azure-slurm-install/install.py
@@ -400,16 +400,16 @@ def _complete_install_primary(s: InstallSettings) -> None:
     ilib.copy_file(
         "imex_prolog.sh",
         f"{s.config_dir}/prolog.d/imex_prolog.sh",
-        owner=s.slurm_user,
-        group=s.slurm_grp,
+        owner="root",
+        group="root",
         mode="0755",
         )
     
     ilib.copy_file(
         "imex_epilog.sh",
         f"{s.config_dir}/epilog.d/imex_epilog.sh",
-        owner=s.slurm_user,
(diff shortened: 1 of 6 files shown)


azure-slurm-install/install.py
```diff
@@ -397,6 +397,22 @@ def _complete_install_primary(s: InstallSettings) -> None:
...
+        f"{s.config_dir}/prolog.d/imex_prolog.sh",
+        owner=s.slurm_user,
+        group=s.slurm_grp,
+        mode="0755",
+        )
+    
+    ilib.copy_file(
+        "imex_epilog.sh",
+        f"{s.config_dir}/epilog.d/imex_epilog.sh",
+        owner=s.slurm_user,
```
aditigaur4, line 411: Can you verify that the script does not need to be root owned? prolog/epilog runs as root. I am not sure if this is correct, but do verify.

azreenz: The scripts have been running normally with this setting but i can change it to root

aditigaur4:
Overall looks good,

Lets add a comment in the template describing the support for slurm.imex.enabled parameter ( as a comment). Perhaps set the parameter in the comment.
# describe what it does
#slurm.imex.enabled = true

A new section in the Readme describing the GB200 support and imex support would also be nice and also document the template parameter.

aditigaur4 (approved):

PR #359, opened 2025-05-08
add gpus flag to srun cli

azreenz:
add optional gpus flag to srun cli to run nvidia-smi commands

### DIFF
--- a/azure-slurm/slurmcc/util.py
+++ b/azure-slurm/slurmcc/util.py
@@ -32,7 +32,7 @@ class NativeSlurmCLI(ABC):
         ...
 
     @abstractmethod
-    def srun(self, hostname: List[str], user_command: str, timeout: int, shell: bool, partition: str) -> SrunOutput:
+    def srun(self, hostname: List[str], user_command: str, timeout: int, shell: bool, partition: str, gpus: int) -> SrunOutput:
         ...
 
 class NativeSlurmCLIImpl(NativeSlurmCLI):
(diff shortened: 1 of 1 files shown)


azure-slurm/slurmcc/util.py
```diff
@@ -43,15 +43,16 @@ def scontrol(self, args: List[str], retry: bool = True) -> str:
...
         with tempfile.NamedTemporaryFile(delete=True) as temp_file:
             temp_file_path = temp_file.name
 
             try:
                 command = f"bash -c '{user_command}'" if shell else user_command
                 partition_flag = f"-p {partition} " if partition else ""
+                gpu_flag = f"--gpus={gpus} " if gpus else ""
                 #adding deadline timeout 1 minute more than the srun timeout to avoid deadline timeout before srun can finish running
-                srun_command = f"srun {partition_flag}-w {','.join(hostlist)} --error {temp_file_path} --deadline=now+{timeout+1}minute --time={timeout} {command}"
+                srun_command = f"srun {partition_flag}-w {','.join(hostlist)} {gpu_flag}--error {temp_file_path} --deadline=now+{timeout+1}minute --time={timeout} {command}"
```
aditigaur4, line 55: I mean the way this is written is not functionally wrong, but there are better ways to write it. For example:
```
args = []
args.append("srun")
if partition_flag:
    args.append(f"-p {partition}")
if gpu_flag:
    args.append(f"--gpus={gpus}")
args.append(f"-w  {','.join(hostlist)}")
args.append(f"--error {temp_file_path}")
...
```
and then once you have assembled it, its just as simple as:
```
cmd = " ".join(args)
```
The reason this cleaner is that list already keeps things in order in which they were inserted. Making future changes just requires adding things to a list,
Whereas in your case you have to carefully manipulate a string on line 55 and ensure the none of your arguments violate any spacing etc. Already the spacing looks a little non-intuitive because you have baked the logic of adding spacing into the arguments `partition_flag = f"-p {partition} "`. Which means everytime you add an argument you need to remember to add a space with it.

azreenz: This makes sense, thanks for the suggestion. Added this change and tested it on a cluster

aditigaur4 (approved):

PR #380, opened 2025-07-08
Do dns lookup for each host for imex_prolog scripts

azreenz:


aditigaur4:
>  Let's get this working, then do a 2nd iteration were we regexp for non-IPs and only nslookup those.

What do you mean by this? I think we are only looking up nodes in a job list, is there a way to condense it further?
Agreed that nslookup is non-ideal but its faster than reading IP addresses via scontrol (which sends RPC to the controller) and is not recommended by schedMD plus we have an outstanding item on @ryan's docket to remove IP addresses.

anhoward:
We shouldn't be using `nslookup`, we need to use `getent hosts`. `nslookup` doesn't follow the full DNS chain.

aditigaur4:
> We shouldn't be using `nslookup`, we need to use `getent hosts`. `nslookup` doesn't follow the full DNS chain.

Hehe! yeah we just made that change

PR #381, opened 2025-07-10
Remove shutdown policy option

anhoward:
"Deallocate" doesn't work properly with Slurm clusters, so removing the option from the default template.


aditigaur4 (approved):
Thanks Andy

PR #394, opened 2025-07-24
update blockname to include ClusterUUID and CliqueID

azreenz:
updating blockname to include clusterUUID_CLiqueID so that azslurmd can parse scontrol show topology using the clusterUUID and CliqueID to detect if nodename is in correct block
Add total # of blocks at the top of the topo
Validate each host has the correct rack id format before adding it to the topo

### DIFF
--- a/azure-slurm/slurmcc/topology.py
+++ b/azure-slurm/slurmcc/topology.py
@@ -204,7 +204,7 @@ class Topology:
                     log.warning("Invalid rack id format: %s, expected format 'ClusterUUID_CliqueID', Skipping host", line)
                     log.warning(f"Error: {host_out.stderr}")
                     break
-                cluster_uuid, clique_id = line.split('_')
+                cluster_uuid, clique_id = line.split('_', 1)
                 if not cluster_uuid or not clique_id:
                     log.warning("Invalid rack id format: %s, expected format 'ClusterUUID_CliqueID', Skipping host", line)
                     log.warning(f"Error: {host_out.stderr}")


aditigaur4:
I am not sure I understand the point of this PR

aditigaur4:
what does that mean:

> so that azslurmd can parse topology.conf using the clusterUUID and CliqueID to detect if nodename is in correct block

azreenz:
> what does that mean:
> 
> > so that azslurmd can parse topology.conf using the clusterUUID and CliqueID to detect if nodename is in correct block

so @ryanhamel  proposed, if i understand correctly, when a new node gets added to the cluster we will update the features using scontrol update nodename =hpc1 feature=clusteruuid_abc,cliqueid_xyz then we wait for azslurmd to update the topo for 3 min, in that 3 min we check to see if hpc1 shows up in the right topology block via scontrol show topology | grep abc_xyz to make sure that hpc1 is in that block and then we start up slurmd. This pr is so that when we do scontrol show topology we can grep for the clusteruuid_cliqueid that we get from the features from scontrol show node hpc1 instead of an arbitrary index.

azreenz:
> > > what does that mean:
> > > > so that azslurmd can parse topology.conf using the clusterUUID and CliqueID to detect if nodename is in correct block
> > 
> > 
> > so @ryanhamel proposed, if i understand correctly, when a new node gets added to the cluster we will update the features using scontrol update nodename =hpc1 feature=clusteruuid_abc,cliqueid_xyz then we wait for azslurmd to update the topo for 3 min, in that 3 min we check to see if hpc1 shows up in the right topology block via scontrol show topology | grep abc_xyz to make sure that hpc1 is in that block and then we start up slurmd. This pr is so that when we do scontrol show topology we can grep for the clusteruuid_cliqueid that we get from the features from scontrol show node hpc1 instead of an arbitrary index.
> 
> okay so you just want a way to conveniently edit the topology.conf and to detect which node is in which block. That makes sense from this PR's point of view.

yep exactly so that blockname can give us more information besides an index

PR #397, opened 2025-07-28
set gpu features before setup slurmd

azreenz:
Get ClusterUUID and CliqueId from nvidia-smi and set slurm node features to include gpu info before setup_slurmd. Set node state to drain with reason cyclecloud_waiting_for_topology

### DIFF
--- a/azure-slurm-install/install.py
+++ b/azure-slurm-install/install.py
@@ -638,6 +638,81 @@ def get_gres_count(hostname):
 
     return count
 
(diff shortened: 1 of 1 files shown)


aditigaur4:
FYI this PR obviously should stay draft since it is incomplete. Unlike other PR's this one cannot go in until we validate the auto reconfig logic for topology and that logic (calling appropriate topology regenerate functions) from the right place is absolutely essential because this PR marks the slurm node drain.

PR #408, opened 2025-08-18
Bump default slurm version to 25.05.2

azreenz:
Bump default slurm version to 25.05.2 and only support one version

Add slurm.insiders flag to template. if set to true pull from insiders repo else pull from stable repo

### DIFF
--- a/azure-slurm-install/rhel.sh
+++ b/azure-slurm-install/rhel.sh
@@ -30,7 +30,7 @@ slurm_packages="slurm slurm-libpmi slurm-devel slurm-pam_slurm slurm-perlapi slu
 sched_packages="slurm-slurmctld slurm-slurmdbd slurm-slurmrestd"
 execute_packages="slurm-slurmd"
 
-INSIDERS=$(/opt/cycle/jetpack/bin/jetpack config slurm.insiders null)
+INSIDERS=$(/opt/cycle/jetpack/bin/jetpack config slurm.insiders False)
 
 if [[ "$INSIDERS" == "True" ]]; then
     if [ "$OS_VERSION" -gt "8" ]; then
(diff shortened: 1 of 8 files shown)


azure-slurm-install/rhel.sh
```diff
@@ -30,11 +30,20 @@ slurm_packages="slurm slurm-libpmi slurm-devel slurm-pam_slurm slurm-perlapi slu
...
+INSIDERS=$(/opt/cycle/jetpack/bin/jetpack config slurm.insiders False)
 
-if [ "$OS_VERSION" -gt "8" ]; then
-    cp slurmel9.repo /etc/yum.repos.d/slurm.repo
+if [[ "$INSIDERS" == "True" ]]; then
+    if [ "$OS_VERSION" -gt "8" ]; then
+        cp slurmel9insiders.repo /etc/yum.repos.d/slurm.repo
+    else
+        cp slurmel8insiders.repo /etc/yum.repos.d/slurm.repo
+    fi
```
aditigaur4, line 40: Since we are touching this-- could we improve this check to be more precise? so if the OS_VERSION is 9 then use 9 if its 8 then use 8 else exit with an error

azure-slurm-install/rhel.sh
```diff
@@ -30,11 +30,20 @@ slurm_packages="slurm slurm-libpmi slurm-devel slurm-pam_slurm slurm-perlapi slu
...
 else
-    cp slurmel8.repo /etc/yum.repos.d/slurm.repo
+    if [ "$OS_VERSION" -gt "8" ]; then
```
aditigaur4, line 42: Same here as well

aditigaur4 (approved):

PR #411, opened 2025-08-27
Remove slurm version suffix from project and install latest version suffix for rhel and ubuntu

azreenz:
remove dependency on full slurm version inlcuding version suffix and pull latest for slurm packages.


azreenz:
Tested on ubuntu24 amd64 and arm64 images, ubuntu22.04 image, and almalinux8 image. All work
Almalinux 9 has powertools repo bug so cant validate fully but change to get latest does work
Rhel 9 doesnt converge because of known issue but rhel.sh runs and installs slurm as expected

aditigaur4 (approved):

PR #416, opened 2025-09-25
move slurm_exporter to cyclecloud-slurm project

azreenz:
Move slurm_exporter from cyclecloud-monitoring to cyclecloud-slurm project.
This PR prevents cyclecloud-monitoring project to prematurely start slurm services before cyclecloud-slurm project is configured and installed properly.

Proposed solution is starting slurmrestd and slurm_exporter in start-services.sh after starting slurmctld and slurmdbd on scheduler node

start-services.sh now includes starting slumrestd.service if jetpack config var slurm.accounting.enabled=true. (slurmrestd is defaulted to start if accounting is enabled).
start-services.sh also starts slurm_exporter if jetpack config var monitoring.enabled=true (This is a workaround solution for now need to discuss further where we should place the slurm_exporter scripts)

created exporter directory in azure-slurm-install to place slurm_exporter.yml file

slurm_exporter and slurmrestd dependency packages have been placed in rhel.sh and ubuntu.sh.

setup_slurmrestd and configure_JWT_authentication functions have been added to install.py


azreenz:
Tested on Ubuntu 22 image with monitoring.enabled and slurm.accounting.enabled flag set to true and slurmrestd  (Ubuntu 24, almalinux, and rhel image scheduler does not converge for azslurm 4.0.3)
Still need to test with monitoring project

azure-slurm-install/exporter/60_slurm_exporter.sh
```diff
@@ -0,0 +1,156 @@
...
+fi
+
+echo "Installing Slurm Exporter..."
+
+install_prerequisites() {
+    # Configure JWT
+
+    # Create a local key
+    mkdir -pv /var/spool/slurm/statesave
+    dd if=/dev/random of=/var/spool/slurm/statesave/jwt_hs256.key bs=32 count=1
```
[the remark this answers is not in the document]
azreenz: could change if needed, copied these lines from cyclecloud-monitoring project

azure-slurm-install/exporter/60_slurm_exporter.sh
```diff
@@ -0,0 +1,156 @@
+#!/bin/bash
+set -e
+script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+PROM_CONFIG=/opt/prometheus/prometheus.yml
+
+source "$script_dir/common.sh"
+
+SLURM_EXPORTER_PORT=9080
+
+SLURM_EXPORTER_REPO="https://github.com/SlinkyProject/slurm-exporter.git"
```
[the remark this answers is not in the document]
azreenz: wont fix

azure-slurm-install/exporter/60_slurm_exporter.sh
```diff
@@ -0,0 +1,156 @@
...
+    lines_to_insert="AuthAltTypes=auth/jwt\nAuthAltParameters=jwt_key=/var/spool/slurm/statesave/jwt_hs256.key\n"
+    if ! grep -q "AuthAltTypes=auth/jwt" /etc/slurm/slurm.conf; then
+        sed -i --follow-symlinks '/^Include azure.conf/a '"$lines_to_insert"'' /etc/slurm/slurm.conf
+    fi
+    # if /etc/slurm/slurmdbd.conf exists then insert the JWT token
+    if [ -e /etc/slurm/slurmdbd.conf ]; then
+        if ! grep -q "AuthAltTypes=auth/jwt" /etc/slurm/slurmdbd.conf; then
+            sed -i --follow-symlinks '/^# Authentication info/a '"$lines_to_insert"'' /etc/slurm/slurmdbd.conf
+        fi
+    fi
```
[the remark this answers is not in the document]
azreenz: could change if needed, copied these lines from cyclecloud-monitoring project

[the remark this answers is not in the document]
azreenz: could change if needed, copied these lines from cyclecloud-monitoring project

azure-slurm-install/exporter/common.sh
```diff
@@ -0,0 +1,44 @@
...
+
+function is_arm64() {
+   uname -m | grep -q 'aarch64\|arm64' 
+}
+
+
+
+function install_yq() {
+    # Install yq
+    YQ_VERSION="v4.44.2"
```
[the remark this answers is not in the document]
azreenz: could change if needed, copied these lines from cyclecloud-monitoring project

azure-slurm-install/start-services.sh
```diff
@@ -55,11 +55,66 @@ for i in $( seq 1 $attempts ); do
...
+SLURMRESTD_OPTIONS="-u slurmrestd -g slurmrestd"
+SLURMRESTD_LISTEN=:6820,unix:/var/spool/slurmrestd/slurmrestd.socket
+EOF
+    chmod 644 /etc/default/slurmrestd
+
+    # Add required capabilities to systemd unit
+    mkdir -p /etc/systemd/system/slurmrestd.service.d
+    cat <<EOF > /etc/systemd/system/slurmrestd.service.d/override.conf
+[Service]
+AmbientCapabilities=CAP_SETGID
```
azreenz, line 100: added these lines because of an CAPSETGID error i was getting when trying to run slurmrestd 
```
× slurmrestd.service - Slurm REST daemon
     Loaded: loaded (/lib/systemd/system/slurmrestd.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Mon 2025-09-29 16:19:37 UTC; 3ms ago
    Process: 15769 ExecStart=/usr/sbin/slurmrestd $SLURMRESTD_OPTIONS (code=exited, status=1/FAILURE)
   Main PID: 15769 (code=exited, status=1/FAILURE)
        CPU: 3ms

Sep 29 16:19:37 master-925-scheduler systemd[1]: Started Slurm REST daemon.
Sep 29 16:19:37 master-925-scheduler slurmrestd[15769]: [2025-09-29T16:19:37.350] fatal: slurmrestd process lacks CAP_SETGID to drop supplementary groups. Supplementary groups must be >
Sep 29 16:19:37 master-925-scheduler systemd[1]: slurmrestd.service: Main process exited, code=exited, status=1/FAILURE
Sep 29 16:19:37 master-925-scheduler systemd[1]: slurmrestd.service: Failed with result 'exit-code'.
```

aditigaur4 (changes requested):
This PR needs to be mostly re-done. We follow a process in installlib/install.py to install slurmd slurmctld. Why are we not following the same convention to install and setup slurmrestd?

azure-slurm-install/start-services.sh
aditigaur4: One thing i did want to mention is that since we are launching a docker container for the exporter-- we should treat it like packages we install in rhel.sh/ubuntu.sh. So if someone wants to just install all our stuff (lets say to make custom images) we install the packages along with downloading the container (container should only be downloaded on the scheduler node). So make sure rhel.sh/ubuntu.sh also downloads the container.

azure-slurm-install/start-services.sh
```diff
@@ -55,11 +55,31 @@ for i in $( seq 1 $attempts ); do
...
+slurmrestd_disabled=$(/opt/cycle/jetpack/bin/jetpack config slurmrestd.disabled False)
+accounting_enabled=$(/opt/cycle/jetpack/bin/jetpack config slurm.accounting.enabled False)
+if [[ "$slurmrestd_disabled" == "False" && "$accounting_enabled" == "True" ]]; then
+    sleep 10
+    systemctl start slurmrestd
+    systemctl status slurmrestd
+    if [ $? != 0 ]; then
+        echo Warning: slurmrestd failed to start! 1>&2
+        echo Here are the last 100 lines of slurmrestd.log
+        tail -n 100 /var/log/slurmctld/slurmrestd.log 1>&2
```
azreenz, line 77: not sure actually how to configure this log... will look into it further or remove these lines

azure-slurm-install/exporter/60_slurm_exporter.sh
```diff
@@ -0,0 +1,126 @@
...
+
+run_slurm_exporter() {
+
+    # Run Slurm Exporter in a container
+    unset SLURM_JWT
+    export $(scontrol token username="slurmrestd" lifespan=infinite)
+    # Check if the token is set
+    if [ -z "$SLURM_JWT" ]; then
+        echo "Failed to get SLURM_JWT token - restarting slurm"
+        systemctl restart slurmctld
```
azreenz, line 65: whats the best way to restart slurmctld here if SLURM_JWT token is not retrieved? Previous script ran the start-services script but now start-services script is calling this script

azure-slurm-install/start-services.sh
```diff
@@ -55,11 +55,31 @@ for i in $( seq 1 $attempts ); do
...
+    echo Here are the last 100 lines of slurmctld.log
+    tail -n 100 /var/log/slurmctld/slurmctld.log 1>&2
+    exit 2
+fi
 
-echo FATAL: slurmctld has not started! 1>&2
-echo Here are the last 100 lines of slurmctld.log
-tail -n 100 /var/log/slurmctld/slurmctld.log 1>&2
-exit 2
+slurmrestd_disabled=$(/opt/cycle/jetpack/bin/jetpack config slurmrestd.disabled False)
```
aditigaur4, line 68: Whats the purpose of having slurmrestd_disabled? Is it needed?

azure-slurm-install/exporter/60_slurm_exporter.sh
aditigaur4: Previous implementation was very buggy and actually never worked correctly in production. So dont copy what it did only use it as a reference. First remove unnecessary scripts and put everything in 1 place. Secondly-- i would look into documentation as to how jwt token can fail. Perhaps just retrying the token call 5 times or so is okay. Does restarting slurmctld actually help? I doubt it.

azure-slurm-install/start-services.sh
```diff
@@ -55,11 +55,31 @@ for i in $( seq 1 $attempts ); do
...
+    systemctl status slurmrestd
+    if [ $? != 0 ]; then
+        echo Warning: slurmrestd failed to start! 1>&2
+        echo Here are the last 100 lines of slurmrestd.log
+        tail -n 100 /var/log/slurmctld/slurmrestd.log 1>&2
+        exit 2
+    fi
+    monitoring_enabled=$(/opt/cycle/jetpack/bin/jetpack config monitoring.enabled False)
+    if [[ "$monitoring_enabled" == "True" ]]; then
+        . "${script_dir}/exporter/60_slurm_exporter.sh"
```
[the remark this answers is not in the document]
azreenz: i didnt wanna clutter up start-services.sh at first since the exporter script i got from monitoring project had a bunch of functions. now it has been cleaned up so should i just add the two functions run_slurm_exporter and add_scraper directly into start-services?

azure-slurm-install/start-services.sh
azreenz: yeah I agree it is getting confusing. I guess we can keep it to if accounting is enabled then slurmrestd will automatically start and slurm_exporter will only start if monitoring is enabled

azure-slurm-install/start-services.sh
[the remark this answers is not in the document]
aditigaur4: Well you need to add logic in start services that

1) starts whatever services needed to make this work.

thats your end goal here. How you achieve it (whether to copy this function or that is irrelevant, esp since the code you are copying from is actually buggy) is by investigating what you are doing and how it ties into exporting the metrics from slurm. Do you need to restart prometheus every time a scraper is added? Or was that just something someone else did? Can adding config to prometheus scraper be added in the configuration section?


The reason we are doing this is because we need to have a 3-fold separation:

1) installation -- download all packages, dependencies (rhel.sh, ubuntu.sh)
2) configuration--- add any config files etc  (install.py)
3) start services as the final step. (start-services.sh)

The reason we want this separation is so that 3 steps can be run independently. For example if someone wants to build custom images they can run installation seperately. Similarly if someone has built a custom image (all required dependencies are present, then we dont need to hit the internet) then we can just do configuration.

Start_services is the final step always.

So the question is not what you need to copy where. I would prefer you investigate each and every addition and determine if its correct and needed and put pieces into appropriate buckets as defined above.

azure-slurm-install/start-services.sh
```diff
@@ -55,11 +55,101 @@ for i in $( seq 1 $attempts ); do
...
+
+    # Check if the container is running
+    if [ "$(docker ps -q -f ancestor=$SLURM_EXPORTER_IMAGE_NAME)" ]; then
+        echo "Slurm Exporter is running"
+    else
+        echo "Slurm Exporter is not running"
+        /opt/cycle/jetpack/bin/jetpack log "Slurm Exporter container failed to start" --level=warn --priority=medium
+        return 0 # do not fail the slurm startup if exporter fails
+    fi
+    #add scraper if prometheus is installed
```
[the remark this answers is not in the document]
azreenz: will verify

azure-slurm-install/start-services.sh
```diff
@@ -55,11 +55,101 @@ for i in $( seq 1 $attempts ); do
...
+    done
 
-echo FATAL: slurmctld has not started! 1>&2
-echo Here are the last 100 lines of slurmctld.log
-tail -n 100 /var/log/slurmctld/slurmctld.log 1>&2
-exit 2
+    if [ -z "$SLURM_JWT" ]; then
+        echo "Failed to get SLURM_JWT token after 3 attempts."
+        echo "Check slurmctld status, slurm.conf JWT configuration, and logs for errors."
+        /opt/cycle/jetpack/bin/jetpack log "Failed to get SLURM_JWT token after 3 attempts, disabling slurm_exporter setup." --level=warn --priority=medium
```
azreenz, line 91: made priority medium so it shows up on the CC UI is this adequate or do we want it to only be in the logs?

azure-slurm-install/install.py
```diff
@@ -445,7 +475,11 @@ def _complete_install_primary(s: InstallSettings) -> None:
             "epilog": "/etc/slurm/epilog.d/*",
             "launch_parameters" : s.launch_parameters,
             "health_interval": health_interval,
-            "health_program": health_program
+            "health_program": health_program,
+            "auth_alt_type": (
+                "AuthAltTypes=auth/jwt\nAuthAltParameters=jwt_key=/var/spool/slurm/statesave/jwt_hs256.key\n"
+                if s.monitoring_enabled else ""
```
azreenz, line 481: should I add # isntead of empty line here?

azure-slurm-install/start-services.sh
```diff
@@ -55,11 +55,110 @@ for i in $( seq 1 $attempts ); do
...
+    fi
+    INSTANCE_NAME=$(hostname)
+
+    yq eval-all '. as $item ireduce ({}; . *+ $item)' $PROM_CONFIG $script_dir/exporter/slurm_exporter.yml > tmp.yml
+    mv -vf tmp.yml $PROM_CONFIG
+
+    # update the configuration file
+    sed -i "s/instance_name/$INSTANCE_NAME/g" $PROM_CONFIG
+
+    # Find the Prometheus process and send SIGHUP to reload config
```
azreenz, line 129: confirmed i can send SIGHUP to prometheus process to reload config instead of restarting service

azure-slurm-install/start-services.sh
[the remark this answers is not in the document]
azreenz: monitoring project gets installed after slurm project so i am unable to add slurm_exporter scraper to prometheus until monitoring project gets set up. so i dont think i can add the scraper stuff into install.py

azure-slurm-install/start-services.sh
azreenz: keep this sighup here? to reload config file after slurm_exporter starts

azure-slurm-install/install.py
```diff
@@ -713,6 +746,78 @@ def setup_slurmd(s: InstallSettings) -> None:
     )
     ilib.enable_service("slurmd")
 
+def setup_slurmrestd(s: InstallSettings) -> None:
```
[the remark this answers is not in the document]
azreenz: look into this further, could be identical to how slurmdbd, maybe slurmrestd can be configured for both scheduler nodes

azure-slurm-install/start-services.sh
```diff
@@ -55,11 +55,101 @@ for i in $( seq 1 $attempts ); do
...
+
+#start slurmrestd if accounting is enabled
+accounting_enabled=$(/opt/cycle/jetpack/bin/jetpack config slurm.accounting.enabled False)
+if [[ "$accounting_enabled" == "True" ]]; then
+    sleep 10
+    systemctl start slurmrestd
+    systemctl status slurmrestd
+    if [ $? != 0 ]; then
+        echo Warning: slurmrestd failed to start! 1>&2
+        exit 2
```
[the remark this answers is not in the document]
azreenz: yes keep as error

azure-slurm-install/start-services.sh
azreenz: we can perhaps reload config before slurm exporter officially starts but that may cause issues, will verify

azure-slurm-install/start-services.sh
azreenz: Ben said we can get rid of slurmrestd.disabled as well as get rid of the dependency for slurmrestd to start if accounting is enabled. slurmrestd will default to start always (config will change based on if accounting is enabled or not) and slurm_exporter will only start if monitoring is enabled so it will reduce the gating logic

azure-slurm-install/start-services.sh
```diff
@@ -55,11 +55,90 @@ for i in $( seq 1 $attempts ); do
...
+    if [ "$(docker ps -q -f ancestor=$SLURM_EXPORTER_IMAGE_NAME)" ]; then
+        echo "Slurm Exporter is running"
+    else
+        echo "Slurm Exporter is not running"
+        /opt/cycle/jetpack/bin/jetpack log "Slurm Exporter container failed to start" --level=warn --priority=medium
+        return 0 # do not fail the slurm startup if exporter fails
+    fi
+
+    # Find the Prometheus process and send SIGHUP to reload config or log a warning if not found
+    PROM_PID=$(pgrep -f 'prometheus')
```
azreenz, line 113: i think it makes sense to me to keep this part (reloading prometheus config if prometheus is running) in start-services after we run slurm_exporter container so that we only reload config at the end of all the install and configuration.

azure-slurm-install/templates/slurm_exporter.yml
```diff
@@ -0,0 +1,9 @@
+scrape_configs:
+  - job_name: slurm_exporter
+    static_configs:
+      - targets: ["instance_name:9080"]
+    relabel_configs:
```
azreenz, line 5: moved this file to templates/ dir since it is called by install.py, is this ok?

azure-slurm-install/install.py
```diff
@@ -445,7 +476,11 @@ def _complete_install_primary(s: InstallSettings) -> None:
             "epilog": "/etc/slurm/epilog.d/*",
             "launch_parameters" : s.launch_parameters,
             "health_interval": health_interval,
-            "health_program": health_program
+            "health_program": health_program,
+            "auth_alt_type": (
+                "AuthAltTypes=auth/jwt\nAuthAltParameters=jwt_key=/var/spool/slurm/statesave/jwt_hs256.key\n"
```
aditigaur4, line 481: I think the path to jwt_key can be added as a member of the installsettings class since its mentioned in a bunch of places

azure-slurm-install/install.py
```diff
@@ -713,6 +746,119 @@ def setup_slurmd(s: InstallSettings) -> None:
     )
     ilib.enable_service("slurmd")
 
+def setup_slurmrestd(s: InstallSettings) -> None:
+    if s.mode != "scheduler":
+        logging.info("Running on non-scheduler node skipping this step.")
+        return
+        
+    # Add slurmrestd to docker group
```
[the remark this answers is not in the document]
aditigaur4: i think its fine as long as it works.

azure-slurm-install/install.py
aditigaur4, line 481: `AuthAltParameters`, should be a new key value pair. Dont just add it as a string with a newline in between. Which means you need to add it to the slurm.conf.template too.

azure-slurm-install/start-services.sh
azreenz: yeah it does it gets configured in monitoring project

azure-slurm-install/install.py
```diff
@@ -713,6 +746,119 @@ def setup_slurmd(s: InstallSettings) -> None:
...
+        ilib.file(jwt_key, content=key, owner=s.slurm_user, group=s.slurm_grp, mode=600)
+    else:
+        ilib.chown(jwt_key, owner=s.slurm_user, group=s.slurm_grp)
+        ilib.chmod(jwt_key, mode=600)
+
+    ilib.chown(jwt_dir, owner=s.slurm_user, group=s.slurm_grp)
+    ilib.chmod(jwt_dir, mode=755)
+    ilib.chmod("/var/spool/slurm", mode=755)
+
+def _add_slurm_exporter_scraper(s: InstallSettings, prom_config: str, exporter_yaml: str) -> None:
```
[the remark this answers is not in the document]
aditigaur4: To validate you can temporarily add monitoring project in the template before the slurm project. Look into how healthagent is added to slurm template, you ll have to follow similarly.

azure-slurm-install/install.py
azreenz: gotcha i'll fix that

aditigaur4:
Overall looks good, just a few minor changes. But overall its looking pretty good.

azure-slurm-install/install.py
azreenz: the path to the key correct?

aditigaur4: yes the whole path to the file. `/var/spool/slurm/statesave/jwt_hs256.key`

aditigaur4 (approved):

PR #440, opened 2025-11-21
Fix suse support

azreenz:
install python311 for suse in 00-install.sh
disable healthchecks, slurmrestd, monitoring, and enroot/pyxis support for suse


azreenz:
I added isolated suse checks so i dont mess with working logic and accidently create bugs....

aditigaur4 (approved):

PR #488, opened 2026-03-18
Azslurm exporter integration

azreenz:
Package, install and start azslurm-exporter when monitoring is enabled. Remove old slurm-exporter integration

### DIFF
--- a/azure-slurm-install/start-services.sh
+++ b/azure-slurm-install/start-services.sh
@@ -117,17 +117,13 @@ run_slurmrestd() {
         echo Warning: slurmrestd is not supported on SUSE, skipping start. 1>&2
         exit 0
     fi
-    monitoring_enabled=$(/opt/cycle/jetpack/bin/jetpack config cyclecloud.monitoring.enabled False)
+
     systemctl start slurmrestd
     systemctl status slurmrestd --no-pager > /dev/null
     if [ $? != 0 ]; then
         echo Warning: slurmrestd failed to start! 1>&2
         /opt/cycle/jetpack/bin/jetpack log "slurmrestd failed to start" --level=warn --priority=medium
         exit 0
-    fi
-    # start slurm_exporter if monitoring is enabled and slurmrestd is running
-    if [[ "$monitoring_enabled" == "True" ]]; then
-        run_slurm_exporter
     fi
 }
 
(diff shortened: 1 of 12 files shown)


azure-slurm-install/start-services.sh
```diff
@@ -146,67 +142,37 @@ reload_prom_config(){
...
-    if [ -z "$SLURM_JWT" ]; then
-        echo "Failed to get SLURM_JWT token after 3 attempts."
-        echo "Check slurmctld status, slurm.conf JWT configuration, and logs for errors."
-        /opt/cycle/jetpack/bin/jetpack log "Failed to get SLURM_JWT token after 3 attempts, disabling slurm_exporter setup." --level=warn --priority=medium
+    systemctl start azslurm-exporter
+    systemctl status azslurm-exporter --no-pager > /dev/null
+    if [ $? != 0 ]; then
+        echo Warning: azslurm-exporter failed to start! 1>&2
+        /opt/cycle/jetpack/bin/jetpack log "azslurm-exporter failed to start" --level=warn --priority=medium
         return 0
```
azreenz, line 165: if monitoring is enabled and azslurm-exporter doesnt start should we exit out of script or keep it as return 0?

specs/scheduler/cluster-init/scripts/10-install_azslurm_exporter.sh
```diff
@@ -0,0 +1,97 @@
...
+        echo "Creating virtual environment at $VENV..."
+        $PYTHON_PATH -m venv "$VENV"
+    else
+        echo "Virtual environment already exists at $VENV"
+    fi
+
+    set +e
+    source $VENV/bin/activate
+    set -e
+
```
aditigaur4: Is the goal here to use azslurm venv? or create a separate venv for exporter?

azreenz: Ooh, i am creating a separate venv for exporter but we prob could just use the azslurm venv here. I can update that

azreenz: Since the assumption is 00-install.sh will succeed before 10-install_azslurm_exporter.sh is executed we can just use that venv directly right? we dont have to find the python path and what not. so just activate the exisiting azslurm venv and install the exporter

specs/scheduler/cluster-init/files/common.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
+set -e
+
```
aditigaur4, line 3: As a potential future improvement try to clearly define a MINIMUM_PYTHON_VERSION...and compare python to that. but this is okay for this PR

aditigaur4 (approved):

PR #504, opened 2026-04-28
default Slurm Accounting SSL certificate to be combined certs required for Azure MySQL Database

azreenz:
Always install three SSL certificatetes and combine them to AzureCA{version}.pem as advised by Azure during root certificate rotation. https://learn.microsoft.com/en-us/azure/mysql/flexible-server/security-tls-root-certificate-rotation#steps
Additionally, install optional custom SSL cert slurm.accounting.certificate specified by user in template

The three certificates installed are 

- https://cacerts.digicert.com/DigiCertGlobalRootCA.crt.pem
- https://cacerts.digicert.com/DigiCertGlobalRootG2.crt.pem
- https://www.microsoft.com/pkiops/certs/Microsoft%20RSA%20Root%20Certificate%20Authority%202017.crt

### DIFF
--- a/azure-slurm-install/install.py
+++ b/azure-slurm-install/install.py
@@ -9,9 +9,8 @@ import sys
 import installlib as ilib
 from typing import Dict, Optional
 
-# Legacy: used for connecting to Azure MariaDB, which is deprecated.
-LOCAL_AZURE_CA_PEM = "AzureCA.pem"
-
+# Combined Certs needed to connect to Azure Database for MySQL Server
+COMBINED_CERTS = "AzureCA_4.0.8.pem"
 
 class InstallSettings:
     def __init__(self, config: Dict, platform_family: str, mode: str) -> None:
@@ -98,8 +97,9 @@ class InstallSettings:
         self.acct_user: Optional[str] = config["slurm"]["accounting"].get("user")
         self.acct_pass: Optional[str] = config["slurm"]["accounting"].get("password")
         self.acct_url: Optional[str] = config["slurm"]["accounting"].get("url")
-        self.acct_cert_url: Optional[str] = config["slurm"]["accounting"].get("certificate_url")
+        self.custom_acct_cert: Optional[str] = config["slurm"]["accounting"].get("certificate")
         self.acct_storageloc :Optional[str] = config["slurm"]["accounting"].get("storageloc")
+        self.acct_cert_filename = "CustomCA.pem" if self.custom_acct_cert else COMBINED_CERTS
 
         self.use_nodename_as_hostname = config["slurm"].get(
             "use_nodename_as_hostname", False
@@ -393,30 +393,25 @@ AccountingStorageTRES=gres/gpu
 """,
     )
 
-    # Previously this was required when connecting to any Azure MariaDB instance.
-    # Which is why we shipped with LOCAL_AZURE_CA_PEM.
-    if s.acct_cert_url and s.acct_cert_url != LOCAL_AZURE_CA_PEM:
-        logging.info(f"Downloading {s.acct_cert_url} to {s.config_dir}/AzureCA.pem")
-        subprocess.check_call(
-            [
-                "wget",
-                "-O",
-                f"{s.config_dir}/AzureCA.pem",
-                s.acct_cert_url,
-            ]
-        )
-        ilib.chown(
-            f"{s.config_dir}/AzureCA.pem", owner=s.slurm_user, group=s.slurm_grp
-        )
-        ilib.chmod(f"{s.config_dir}/AzureCA.pem", mode="0600")
-    elif s.acct_cert_url and s.acct_cert_url == LOCAL_AZURE_CA_PEM:
-        ilib.copy_file(
-            LOCAL_AZURE_CA_PEM,
-            f"{s.config_dir}/AzureCA.pem",
+    # Always copy the bundled certificate
+    ilib.copy_file(
+        COMBINED_CERTS,
+        f"{s.config_dir}/{COMBINED_CERTS}",
+        owner=s.slurm_user,
+        group=s.slurm_grp,
+        mode="0600",
+    )
+    #Install custom SSL cert if specified
+    if s.custom_acct_cert:
+        cert_path = f"{s.config_dir}/{s.acct_cert_filename}"
+        logging.info(f"Saving custom SSL cert to {cert_path}")
+        ilib.file(
+            cert_path,
             owner=s.slurm_user,
             group=s.slurm_grp,
             mode="0600",
-        )
+            content=s.custom_acct_cert,
+            )
 
     # Configure slurmdbd.conf
     ilib.template(
@@ -430,9 +425,7 @@ AccountingStorageTRES=gres/gpu
             "dbuser": s.acct_user or "root",
             "dbdhost": s.hostname,
             "storagepass": f"StoragePass={s.acct_pass}" if s.acct_pass else "#StoragePass=",
-            "storage_parameters": "StorageParameters=SSL_CA=/etc/slurm/AzureCA.pem"
-                                  if s.acct_cert_url
-                                  else "#StorageParameters=",
+            "storage_parameters": f"StorageParameters=SSL_CA=/etc/slurm/{s.acct_cert_filename}",
             "slurmver": s.slurmver,
             "storageloc": s.acct_storageloc or f"{s.slurm_db_cluster_name}_acct_db",
             "auth_alt_type": "AuthAltTypes=auth/jwt" if s.jwt_available else "",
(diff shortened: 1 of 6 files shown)


azure-slurm-install/install.py
```diff
@@ -393,9 +457,11 @@ def _accounting_primary(s: InstallSettings) -> None:
 """,
     )
 
+    if s.acct_cert_url and s.acct_cert_url == COMBINED_CERTS:
```
bwatrous: I don't think we need the conditional here - we can simply ALWAYs install the bundle.

azure-slurm-install/install.py
```diff
@@ -353,7 +357,67 @@ def munge_key(s: InstallSettings) -> None:
         mode="0600",
     )
 
+def _install_combined_certs(s: InstallSettings) -> None:
+    """
+    Download and combine three SSL certificates required for Azure MySQL:
```
bwatrous: We should consider bundling the certs at package build time instead of doing it at cluster startup.
The bundle should be very stable.

azure-slurm-install/install.py
```diff
@@ -393,9 +395,17 @@ def _accounting_primary(s: InstallSettings) -> None:
...
+            COMBINED_CERTS,
+            f"{s.config_dir}/AzureCA.pem",
+            owner=s.slurm_user,
+            group=s.slurm_grp,
+            mode="0600",
+        )
     # Previously this was required when connecting to any Azure MariaDB instance.
     # Which is why we shipped with LOCAL_AZURE_CA_PEM.
-    if s.acct_cert_url and s.acct_cert_url != LOCAL_AZURE_CA_PEM:
+    elif s.acct_cert_url and s.acct_cert_url != LOCAL_AZURE_CA_PEM:
```
aditigaur4: We need to remove `LOCAL_AZURE_CA_PEM`. Its deprecated

azure-slurm-install/install.py
```diff
@@ -393,9 +392,15 @@ def _accounting_primary(s: InstallSettings) -> None:
...
+    if s.acct_cert_url and s.acct_cert_url == COMBINED_CERTS:
+        ilib.copy_file(
+            COMBINED_CERTS,
+            f"{s.config_dir}/AzureCA.pem",
+            owner=s.slurm_user,
+            group=s.slurm_grp,
+            mode="0600",
+        )
+    elif s.acct_cert_url and s.acct_cert_url != COMBINED_CERTS:
         logging.info(f"Downloading {s.acct_cert_url} to {s.config_dir}/AzureCA.pem")
```
bwatrous: This feels like a lie.
Why not leave the pem named whatever it is named?

azure-slurm-install/install.py
bwatrous: I would ALWAYS create AzureCA.pem in case a user changes their mind

azure-slurm-install/install.py
azreenz: we normalize it to AzureCA.pem so we can just set /etc/slurm/AzureCA.pem as the SSL_CA path in slurmdbd.conf regardless of what cert we install

bwatrous (dismissed):

azure-slurm-install/install.py
```diff
@@ -458,16 +466,20 @@ def _accounting_all(s: InstallSettings) -> None:
     """
     Perform linking and enabling of slurmdbd
     """
-    # This used to be required for all installations, but it is
-    # now optional, so only create the link if required.
-    original_azure_ca_pem = f"{s.config_dir}/AzureCA.pem"
     ilib.link(
```
aditigaur4, line 455: Whats the point of calling `ilib.link` twice? arent you just overwriting it at the bottom?

You have already have the cert filename-- just set `cert_filename` to `COMBINED_CERTS` if user has not requested a cert.

```
if s.acct_cert_url and s.acct_cert_url != COMBINED_CERTS:
     cert_filename = os.path.basename(urlparse(s.acct_cert_url).path)
else:
    cert_filename = COMBINED_CERTS

ilib.link(...
```

azreenz: Since we are downloading AzureCA_bundle.pem to the cluster regardless of the SSL URL i thought we should link it to /etc/slurm/AzureCA_bundle.pem regardless as well. So if user specifies a url /etc/slurm/ directory will have AzureCA_bundle.pem in addition to their cert

azreenz: so if user wants to change to azureCA_bundle.pem they just change slurmdbd.conf to /etc/slurm/AzureCA_bundle.pem in slurmdbd.conf since we already linked it

aditigaur4: Okay 1) document it (in code) more explicitly. and second-- I find it not very good that this check is repeated 3 times in this diff... 
```
if s.acct_cert_url and s.acct_cert_url != COMBINED_CERTS:
```

templates/slurm.txt
```diff
@@ -801,13 +801,13 @@ Order = 20
...
-        Description = URL to fetch SSL certificate for authentication to DB. Use AzureCA.pem (embedded) for use with deprecated MariaDB instances.
+        Description = URL to fetch SSL certificate for authentication to DB. Default Value installs combined certs needed for Azure Database for MySQL.
         Conditions.Excluded := configuration_slurm_accounting_enabled isnt true
         ParameterType = StringList
         Config.Plugin = pico.form.Dropdown
         Config.FreeForm = true
-        Config.Entries := {[Value=""], [Value="AzureCA.pem"]}
-        DefaultValue = ""
+        Config.Entries := {[Value=""], [Value="AzureCA_bundle.pem"]}
+        DefaultValue = "AzureCA_bundle.pem"
```
aditigaur4: I understand this was done before-- but this is weird. This is saying "Either its a URL or its a string that actually points to a file we know about"...

IMO just remove the default value.

azure-slurm-install/install.py
aditigaur4: Can't this check be moved at instantiation level in __init__? and just assign a cert variable that points to the cert user wants in the config? thats just an idea...

templates/slurm.txt
azreenz: okay so when this value is empty we should always point StorageParameters=SSL_CA=/etc/slurm/AzureCA_bundle.pem? Right now in the code if you leave acct_cert_url empty then we comment out StorageParameters in slurmdbd.conf

azreenz:
New Template looks like this 
Screenshot 2026-05-01 111342
Screenshot 2026-05-01 111404


And when custom ssl is enabled both certs are avail in /etc/slurm but slurmdbd.conf points to CustomCA.pem
```
root@azdbd-u22-custom-scheduler:/etc/slurm# ls
AzureCA_4.0.8.pem  azure.conf   gres.conf                        plugstack.conf    site_specific.conf  topology.conf
CustomCA.pem       cgroup.conf  job_submit.lua.azurehpc.example  plugstack.conf.d  slurm.conf
accounting.conf    epilog.d     keep_alive.conf                  prolog.d          slurmdbd.conf
```
`StorageParameters=SSL_CA=/etc/slurm/CustomCA.pem`

aditigaur4 (approved):

azure-slurm-install/install.py
```diff
@@ -9,9 +9,8 @@
 import installlib as ilib
 from typing import Dict, Optional
 
-# Legacy: used for connecting to Azure MariaDB, which is deprecated.
-LOCAL_AZURE_CA_PEM = "AzureCA.pem"
-
+# Combined Certs needed to connect to Azure Database for MySQL Server
+COMBINED_CERTS = "AzureCA_4.0.8.pem"
```
bwatrous, line 13: If this is the Project version, then we should get it from the project.ini

aditigaur4: This is the version...not the date

azure-slurm-install/package.py
```diff
@@ -58,6 +59,54 @@ def _add(name: str, path: Optional[str] = None, mode: Optional[int] = None) -> N
...
+            microsoft_pem_path = os.path.join(tmpdir, "MicrosoftRSA2017.pem")
+            subprocess.check_call([
+                "openssl", "x509",
+                "-inform", "DER",
+                "-in", microsoft_der_path,
+                "-out", microsoft_pem_path
+            ])
+            with open(microsoft_pem_path, "r") as f:
+                combined_pem += f.read() + "\n"
+        cert_name = f"AzureCA_{version}.pem"
```
bwatrous, line 104: Where is version set?

azreenz: line 26 in package.py (it was already set before i am using it here)

bwatrous: Maybe we should leave the file name AzureCA.pem and add a comment inside the file:
 \# AzureCA.pem bundle generated: ${DATE}

azreenz: I agree thats probably better so we don't have another line we have to change when bumping the project version.. @aditigaur4 What do you think?

azure-slurm-install/install.py
bwatrous: But I don't think we should include project version in the name here.
Potentially, we could add a creation date inside the file...

azure-slurm-install/package.py
aditigaur4: I am not sure if additional lines can be generated inside PEM files...if they can then yes thats a good approach too. But i think adding version number makes the file represent the version the project shipped with..which is when the file was packaged. So i think we should still keep the version number in the filename

azure-slurm-install/install.py
aditigaur4: I think we can do both. (if we can add date inside PEM file in a valid way that is)

bwatrous (approved):

PR #510, opened 2026-05-27
add sacct_jobs metric to exporter

azreenz:
Add completed jobs metric to exporter as a Gauge similar to how we are exporting failed jobs metrics.

### DIFF
--- a/azure-slurm-exporter/exporter/sacct.py
+++ b/azure-slurm-exporter/exporter/sacct.py
@@ -77,11 +77,9 @@ class Sacct(BaseCollector):
     def parse_output(self, stdout) -> List[Union[Counter,Gauge]]:
         """
         Parse sacct command output and increment terminal jobs metric for each job in the output to describe total number of finished jobs by partition,
-        exit_code, reason, and state as well as return failed and completed jobs gauges describing each job that failed and completed within the time interval of the query.
+        exit_code, reason, and state as well as return terminal jobs gauges describing each job that failed and completed within the time interval of the query.
         """
-        sacct_failed_jobs = Gauge("sacct_failed_jobs","Per-job indicator for failed terminal jobs in the query interval (value is always `1`)",
-                                    ["jobid","jobname","nodelist","partition", "exit_code","reason","state", "starttime","endtime"], registry=None)
-        sacct_completed_jobs = Gauge("sacct_completed_jobs","Per-job indicator for completed terminal jobs in the query interval (value is always `1`)",
+        sacct_jobs = Gauge("sacct_jobs","Per-job indicator for all terminal jobs in the query interval (value is always `1`)",
                                     ["jobid","jobname","nodelist","partition", "exit_code","reason","state", "starttime","endtime"], registry=None)
 
         lines = stdout.decode().strip().splitlines()
@@ -95,31 +93,19 @@ class Sacct(BaseCollector):
                 reason=reason,
                 state=row.state.lower()).inc()
 
-            if row.state.lower() != "completed":
-                sacct_failed_jobs.labels(
-                    jobid=row.jobid,
-                    jobname=row.jobname,
-                    nodelist=row.nodelist,
-                    partition=row.partition,
-                    exit_code=row.exitcode,
-                    reason=reason,
-                    state=row.state.lower(),
-                    starttime=row.start,
-                    endtime=row.end).set(1)
-            else:
-                sacct_completed_jobs.labels(
-                    jobid=row.jobid,
-                    jobname=row.jobname,
-                    nodelist=row.nodelist,
-                    partition=row.partition,
-                    exit_code=row.exitcode,
-                    reason=reason,
-                    state=row.state.lower(),
-                    starttime=row.start,
-                    endtime=row.end).set(1)
+            sacct_jobs.labels(
+                jobid=row.jobid,
+                jobname=row.jobname,
+                nodelist=row.nodelist,
+                partition=row.partition,
+                exit_code=row.exitcode,
+                reason=reason,
+                state=row.state.lower(),
+                starttime=row.start,
+                endtime=row.end).set(1)
 
 
-        return [self.sacct_terminal_jobs, sacct_failed_jobs, sacct_completed_jobs]
+        return [self.sacct_terminal_jobs, sacct_jobs]
 
     async def sacct_query(self) -> None:
         """
--- a/azure-slurm-exporter/exporter/sacct.py
+++ b/azure-slurm-exporter/exporter/sacct.py
@@ -77,9 +77,11 @@ class Sacct(BaseCollector):
     def parse_output(self, stdout) -> List[Union[Counter,Gauge]]:
         """
         Parse sacct command output and increment terminal jobs metric for each job in the output to describe total number of finished jobs by partition,
-        exit_code, reason, and state as well as return failed jobs gauge describing each job that failed within the time interval of the query.
+        exit_code, reason, and state as well as return failed and completed jobs gauges describing each job that failed and completed within the time interval of the query.
         """
-        sacct_failed_jobs = Gauge("sacct_failed_jobs","Number of failed slurm jobs in an interval",
+        sacct_failed_jobs = Gauge("sacct_failed_jobs","Per-job indicator for failed terminal jobs in the query interval (value is always `1`)",
+                                    ["jobid","jobname","nodelist","partition", "exit_code","reason","state", "starttime","endtime"], registry=None)
+        sacct_completed_jobs = Gauge("sacct_completed_jobs","Per-job indicator for completed terminal jobs in the query interval (value is always `1`)",
                                     ["jobid","jobname","nodelist","partition", "exit_code","reason","state", "starttime","endtime"], registry=None)
 
         lines = stdout.decode().strip().splitlines()
@@ -102,11 +104,22 @@ class Sacct(BaseCollector):
                     exit_code=row.exitcode,
                     reason=reason,
                     state=row.state.lower(),
-                    starttime=self.starttime,
-                    endtime=self.endtime).set(1)
+                    starttime=row.start,
+                    endtime=row.end).set(1)
+            else:
+                sacct_completed_jobs.labels(
+                    jobid=row.jobid,
+                    jobname=row.jobname,
+                    nodelist=row.nodelist,
+                    partition=row.partition,
+                    exit_code=row.exitcode,
+                    reason=reason,
+                    state=row.state.lower(),
+                    starttime=row.start,
+                    endtime=row.end).set(1)
 
 
-        return [self.sacct_terminal_jobs, sacct_failed_jobs]
+        return [self.sacct_terminal_jobs, sacct_failed_jobs, sacct_completed_jobs]
 
     async def sacct_query(self) -> None:
         """
--- a/README.md
+++ b/README.md
@@ -478,6 +478,8 @@ Log can be found under `/var/log/azslurm-exporter.log`
 | Metric | Type | Labels | Description |
 |---|---|---|---|
 | `sacct_terminal_jobs_total` | Counter | `partition`, `exit_code`, `reason`, `state`| Cumulative count of completed/failed/cancelled jobs |
+| `sacct_failed_jobs` | Gauge | `jobid`, `jobname`, `nodelist`, `partition`, `exit_code`, `reason`, `state`, `starttime`, `endtime` | Per-job indicator for failed terminal jobs in the query interval (value is always `1`) |
+| `sacct_completed_jobs` | Gauge | `jobid`, `jobname`, `nodelist`, `partition`, `exit_code`, `reason`, `state`, `starttime`, `endtime` | Per-job indicator for completed jobs in the query interval (value is always `1`) |
 
 Terminal states tracked: `completed`, `failed`, `cancelled`, `timeout`, `node_fail`, `preempted`, `out_of_memory`, `deadline`, `boot_fail`. Exit codes are mapped to human-readable reasons (e.g. `137:0` → `SIGKILL - Force killed`).
 
--- a/README.md
+++ b/README.md
@@ -478,8 +478,7 @@ Log can be found under `/var/log/azslurm-exporter.log`
 | Metric | Type | Labels | Description |
 |---|---|---|---|
 | `sacct_terminal_jobs_total` | Counter | `partition`, `exit_code`, `reason`, `state`| Cumulative count of completed/failed/cancelled jobs |
-| `sacct_failed_jobs` | Gauge | `jobid`, `jobname`, `nodelist`, `partition`, `exit_code`, `reason`, `state`, `starttime`, `endtime` | Per-job indicator for failed terminal jobs in the query interval (value is always `1`) |
-| `sacct_completed_jobs` | Gauge | `jobid`, `jobname`, `nodelist`, `partition`, `exit_code`, `reason`, `state`, `starttime`, `endtime` | Per-job indicator for completed jobs in the query interval (value is always `1`) |
+| `sacct_jobs` | Gauge | `jobid`, `jobname`, `nodelist`, `partition`, `exit_code`, `reason`, `state`, `starttime`, `endtime` | Per-job indicator for all terminal jobs in the query interval (value is always `1`) |
 
 Terminal states tracked: `completed`, `failed`, `cancelled`, `timeout`, `node_fail`, `preempted`, `out_of_memory`, `deadline`, `boot_fail`. Exit codes are mapped to human-readable reasons (e.g. `137:0` → `SIGKILL - Force killed`).
 
--- a/azure-slurm-exporter/dashboards/failed-jobs.json
+++ b/azure-slurm-exporter/dashboards/failed-jobs.json
@@ -116,7 +116,7 @@
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "sacct_failed_jobs{cluster=\"$cluster\",partition=~\"$partition\",state!=\"completed\"} offset -5m",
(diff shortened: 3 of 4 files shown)


azure-slurm-exporter/exporter/sacct.py
```diff
@@ -77,9 +77,11 @@ def export_metrics(self) -> List[Union[Counter,Gauge]]:
     def parse_output(self, stdout) -> List[Union[Counter,Gauge]]:
         """
         Parse sacct command output and increment terminal jobs metric for each job in the output to describe total number of finished jobs by partition,
-        exit_code, reason, and state as well as return failed jobs gauge describing each job that failed within the time interval of the query.
+        exit_code, reason, and state as well as return failed and completed jobs gauges describing each job that failed and completed within the time interval of the query.
         """
-        sacct_failed_jobs = Gauge("sacct_failed_jobs","Number of failed slurm jobs in an interval",
+        sacct_failed_jobs = Gauge("sacct_failed_jobs","Per-job indicator for failed terminal jobs in the query interval (value is always `1`)",
```
bwatrous: Do we still need the sacct_failed_jobs?   Does all the info exist in sacct_completed_jobs too?

azreenz: I kept it for backwards compat sake since the first version of the dashboard uses it but if no one is using it we can simplify it to one sacct_jobs metric

aditigaur4:
I am not quite sure I understand the point of this PR. Could you please explain?

azure-slurm-exporter/exporter/sacct.py
aditigaur4: We are exporting terminal jobs and failed jobs. And now a third completed jobs. Why do we need 3? Arent terminal jobs== failed jobs + completed jobs?

azure-slurm-exporter/exporter/sacct.py
```diff
@@ -102,11 +104,22 @@ def parse_output(self, stdout) -> List[Union[Counter,Gauge]]:
...
+                    partition=row.partition,
+                    exit_code=row.exitcode,
+                    reason=reason,
+                    state=row.state.lower(),
+                    starttime=row.start,
+                    endtime=row.end).set(1)
 
 
-        return [self.sacct_terminal_jobs, sacct_failed_jobs]
+        return [self.sacct_terminal_jobs, sacct_failed_jobs, sacct_completed_jobs]
```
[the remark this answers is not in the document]
aditigaur4: This is a pretty important comment. We earlier only restricted this for failed jobs with the assumption that its a smaller set. And in terminal jobs (which was supposed to be the total jobs) didnt we decide not to show nodelist for preventing cardinality explosion? 

Completed jobs = terminal jobs - failed jobs. That can just be a grafana side query can it be not? Why do we need a new export for that?

azure-slurm-exporter/exporter/sacct.py
aditigaur4: Completed jobs can just be-- terminal jobs - failed jobs? which can be computed grafana side entirely. Do we need a new export for that? Arent we exploding cardinality here by including nodelist as a label?

azreenz: terminal jobs is a counter which exports the number of jobs that finished with a certain exit code (minimal job information) counters are stateful so they export at every scrape which is why including more job information like nodelist will explode the cardinality. Counters are helpful here when we want total count of failed/completed for a range. sacct_failed_jobs and sacct_completed_jobs are both Gauges and stateless. They only export the number of jobs that completed in the window of the sacct query and are only exported for the time interval duration until the next sacct query runs. the gauges reset at every parse. Grafana keeps track of all the previous exported data to build the historical panels. So yes, we are exporting each job and its info but we only export it for the duration of the sacct query interval (5 min) It is similar to the squeue_job_nodes_allocated metric that exports every running job for the duration that it is running. Instead we are now exporting the job data for each terminal job. But yes, while we did this for failed jobs which is a smaller set of data (or should be), the completed jobs metric will export more. But since it is not a counter metric we will not be exporting the total amount of jobs that ever completed since the exporter ran at every scrape (which was the cardinality issue we discussed when trying to add more unqiue label data to the terminal jobs counter). 

Ben was suggesting we just export one Gauge family metric (sacct_jobs or something) that includes all the states and grafana can filter them instead. Regardless the total number of metrics remains unchanged in a scrape interval since it will always be the number of jobs that finished in the sacct query time window

aditigaur4: > But yes, while we did this for failed jobs which is a smaller set of data (or should be), the completed jobs metric will export more. But since it is not a counter metric we will not be exporting the total amount of jobs that ever completed since the exporter ran at every scrape (which was the cardinality issue we discussed when trying to add more unqiue label data to the terminal jobs counter).

Are you exporting a row per job? Even if its a gauge? If that is true-- you are essentially replicating mysql in grafana.

Secondly-- The argument about backward compat is not sound. Users don't have a dependency on individual metric we export. They have a dependency perhaps on the version of dashboard. Now the terminal jobs is a counter, that is true-- but can that dashboard be rebuilt by using completed jobs + failed jobs? And you can use summation to calculate terminal jobs? Is that correct? Which would remove the need for terminal jobs.

azreenz: Yes i am exporting a row per job. and yes, i was thinking about the case a user has old version of dashboard that uses sacct_failed_jobs but new version of exporter. So basically their failed jobs panel will not populate because it is using the old metric and exporter doesnt export that metric anymore. Also, I haven't found a good way to make this panel we have right now that shows the totals of completed and failed jobs using gauges because of prometheus scraping the data multiple times (basically count and sum includes duplicates in the calculation). If we get rid of this panel however we get rid of the terminal jobs counter dependency.
image

What is possible is creating a table from the time series from the Gauges using data transformation and then counting the rows of the table. But yes we are replicating mysql here 
Screenshot 2026-06-12 131134

aditigaur4: If you remove node list as a label- can you avoid exporting a row per job? We dont need to see in grafana what can be shown through `sacct`. We are just replicating data and that is of no use.

azreenz: We can export a gauge that just describes # of jobs that completed or failed in that sacct query time window but then we lose specifc job details like job id, node list, etc. @bwatrous what do you think about this? I see a good use case for showing failed jobs details (what we already have) so that users can see the job id and nodelist right from grafana and do some diagnosis but what is the benefit of showing all job info for jobs that completed successfully as well?

aditigaur4: For failed jobs we need drill down. But replicating the entire job table through completed jobs is not of use. What is the problem that dashboard would solve? What is something that dashboard will show that is not trivially accessible through slurm? when you add job id to the label you are essentially dumping the whole table in grafana.

bwatrous: Hmm, 
I think the old "Failed Jobs" table and the pie chart counting completed vs failed jobs could easily be removed - they're not very useful for debugging.

But I think the replication of the job table is exactly what we want....
This table is for grafana users who likely don't use sacct (and may not even know slurm well).   If I'm looking at Grafana, I think that job table is really the main thing i want to see.   (There are other options - in particular, we may be able to have Grafana directly read mySQL or (much less perferred) offer a second monitoring UI with Job history - but this is really problematic in the cloud since we need reliable URLs across VNets and VPNs.)

It's really the rows for Jobs that had issues of any kind that we care most about, but the fact that jobs are completing (and ability to compare which jobs complete vs which fail) is really useful for debugging.

azure-slurm-exporter/exporter/sacct.py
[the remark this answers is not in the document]
bwatrous: We should scale test this, but if prometheus can't handle a few million rows in a table I'd be shocked.
Importantly, this WON'T significantly impact our rate limit budget for metrics since cardinality of jobs is tiny compared to metrics like CPU Usage.    (That's true even in high throughput schedulers like symphony that may be doing thousands of Jobs per minute.)

We certainly may need to look at controlling the retirement of older job data (max rows or max age) to prevent the table from growing endlessly.

azure-slurm-exporter/exporter/sacct.py
aditigaur4: I disagree. Slurm exporter is not a true debugging dashboard because ultimately there is only so much we can debug from `sacct`, `squeue` `sinfo`. 

The purpose of slurm exporter is to show lightweight visualization of "what is going on with my cluster right now". If people have to run 3 or 4 different terminal commands, then grafana dashboard earns its keep. But if the grafana dashboard is just line by line replicating just 1 cli command-- whats the point of the dashboard?

I also disagree that people will use grafana in lieu of sacct because they are "grafana users" and not "slurm users". Because they will quickly run into serious limitations and will have to use slurm (even if we build the best dashboards). That argument does not hold up. Where grafana shines is when you have combine data from multiple CLI commands and understand "what is going on in my cluster right now?". Then we can have 1 dashboard that answers that question.

> I think the old "Failed Jobs" table and the pie chart counting completed vs failed jobs could easily be removed - they're not very useful for debugging.

I think we should keep that. That answers an important question-- what percentage of my jobs failed/completed between time window X and time window Y? If i told you to answer that question through CLI alone-- you wont be able to do that easily (or quickly).

> It's really the rows for Jobs that had issues of any kind that we care most about, but the fact that jobs are completing (and ability to compare which jobs complete vs which fail) is really useful for debugging.

Again-- why are completed jobs useful for debugging? What needs to be debugged in a successfully completed job? Failed jobs need to be debugged. Because they failed. For completed jobs-- all we need is a total count. How many jobs completed in a time window X? We dont need to know too many details like node list etc etc. For failed jobs however we need the details. Right now we are replicating entire mysql for completed jobs and failed jobs. This is not a sound strategy at all. If people care about specific job ID-- they will have to run a CLI command. Grafana can never be a built-in replacement for CLI. That is not purpose of the tool.

Its not about the rate limits (which will be a problem later). Its about using wrong tool for the job. Grafana is a "live metrics monitoring solution". Its not a historical data analysis tools for which dashboards need to be build directly on top of Mysql and not have our intermediary exporter. Exporting each an every row of table is terrible inefficient in large clusters that can run thousands of job a day.

azure-slurm-exporter/exporter/sacct.py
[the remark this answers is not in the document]
aditigaur4: We should not be doing in grafana what is done in or on top  of my mysql. I disagree with your idea that just because we have this dashboard..we should make it a complete debugging solution. Because it cannot be. We should however make per jobs dashboard which i am prototyping. But this needs to be a high level aggregate dashboard.

bwatrous (dismissed):

aditigaur4:
I ll try to summarize my concerns with this PR:

Every time you create a unique label with a metric-- it creates a new active time series in a prometheus backend. (Azure monitor metrics backend). 

When we add things like job id as a label -- every job gets a unique active time series created in the backend. Now this in and of itself is not a problem because there could be other reasons why job_id may be useful. But in this case we add 7 other labels out of which some that meaninglessly explode cardinality are 1) node list. In 500 node cluster you can 2^500 unique node list combinations. This is an unbounded effectively. 2) start time, end time. Time series databases are meant to have the x axis as the time. By adding literal x axis values as label-- we have again exploded cardinality. 

The other disadvantage of this approach is-- every new time series-- immediately becomes *dead*. Never to be used again. Customer pays double the cost-- for my sql, and for a dead active series which climbs unboundedly.

The only rational label choices were exit code and partition. Because those stay bounded. 

Here is a cautionary note from prometheus: https://prometheus.io/docs/practices/naming/

> CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

In fact azure monitor billing may also grow with this change needlessly depending on if extra billing is applied on growing active time series. The dashboards are more likely to degrade in a cluster with high cluster activity. Dev clusters are not enough to simulate this because in prod clusters we are sending metrics from every node.

Let me also state the complete uselessness of this design:

*grafana can directly be connected to mysql.* You can build far better dashboards with direct DB access. Without any replication. Without any extra azure monitor bill.

azreenz:
> Every time you create a unique label with a metric-- it creates a new active time series in a prometheus backend. (Azure monitor metrics backend).
> 
> When we add things like job id as a label -- every job gets a unique active time series created in the backend. Now this in and of itself is not a problem because there could be other reasons why job_id may be useful. But in this case we add 7 other labels out of which some that meaninglessly explode cardinality are 1) node list. In 500 node cluster you can 2^500 unique node list combinations. This is an unbounded effectively. 2) start time, end time. Time series databases are meant to have the x axis as the time. By adding literal x axis values as label-- we have again exploded cardinality.
> The only rational label choices were exit code and partition. Because those stay bounded.
> 
> Here is a cautionary note from prometheus: https://prometheus.io/docs/practices/naming/
> 
> CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

So since we are exporting the unique job id we are only creating a time series for each job. even adding node list and starttime etc wont change that because we wouldn't have a label pairs like this 
```
job_id=1, node_list=hpc-[1-5]
job_id=1, node_list=hpc-[6-7]
```
So # of time series will always be # of jobs that finished in the sacct query window.
> The other disadvantage of this approach is-- every new time series-- immediately becomes _dead_. Never to be used again. Customer pays double the cost-- for my sql, and for a dead active series which climbs unboundedly.

This is a valid concern to an extent. 
>Pricing for [Azure Monitor managed service for Prometheus](https://docs.azure.cn/en-us/azure-monitor/essentials/prometheus-metrics-overview) is based on [data samples ingested](https://docs.azure.cn/en-us/azure-monitor/containers/kubernetes-monitoring-enable) and [query samples processed](https://docs.azure.cn/en-us/azure-monitor/essentials/azure-monitor-workspace-manage#link-a-grafana-workspace). Data is retained for 18 months at no extra charge. 

So we are only ingesting each job time series for 5 min and then we stop exporting that set of jobs and then those time series become stale and are not ingested anymore. but customers are paying for the querying which they are technically paying for already with mysql. 

Active time series are not growing "unbounded" here because we are always bounded by number of jobs that finished in the sacct query window. However for clusters with high activity this can become problematic.  Steady-state active series = (job completion rate per hour) × (12 hour window) since active series are defined by # of time series that got ingested in the last 12 hours. but if we compare this to the # of metrics that get exported per cpu for our gpu skus the total number of jobs that finished in 12 hours may be miniscule.

aditigaur4:
> > Every time you create a unique label with a metric-- it creates a new active time series in a prometheus backend. (Azure monitor metrics backend).
> > When we add things like job id as a label -- every job gets a unique active time series created in the backend. Now this in and of itself is not a problem because there could be other reasons why job_id may be useful. But in this case we add 7 other labels out of which some that meaninglessly explode cardinality are 1) node list. In 500 node cluster you can 2^500 unique node list combinations. This is an unbounded effectively. 2) start time, end time. Time series databases are meant to have the x axis as the time. By adding literal x axis values as label-- we have again exploded cardinality.
> > The only rational label choices were exit code and partition. Because those stay bounded.
> > Here is a cautionary note from prometheus: https://prometheus.io/docs/practices/naming/
> > CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.
> 
> So since we are exporting the unique job id we are only creating a time series for each job. even adding node list and starttime etc wont change that because we wouldn't have a label pairs like this
> 
> ```
> job_id=1, node_list=hpc-[1-5]
> job_id=1, node_list=hpc-[6-7]
> ```
> 

why not? you are adding node list start time end time as labels are you not?

> So # of time series will always be # of jobs that finished in the sacct query window.
> 

Incorrect. Just because you are bounding the record you are sending to the backend from the exporter -- those records still are stored in backend are they not? Every 5 minutes you are sending the jobs that completed-- but all those records over the previous day as an example are still stored in the backend are they not?  Active time series here means all time series until whatever time azure monitor chooses to keep them. So yes this effectively unbounded.  I believe the limit may be 30 days.
> > The other disadvantage of this approach is-- every new time series-- immediately becomes _dead_. Never to be used again. Customer pays double the cost-- for my sql, and for a dead active series which climbs unboundedly.
> 
> This is a valid concern to an extent.
> 
> > Pricing for [Azure Monitor managed service for Prometheus](https://docs.azure.cn/en-us/azure-monitor/essentials/prometheus-metrics-overview) is based on [data samples ingested](https://docs.azure.cn/en-us/azure-monitor/containers/kubernetes-monitoring-enable) and [query samples processed](https://docs.azure.cn/en-us/azure-monitor/essentials/azure-monitor-workspace-manage#link-a-grafana-workspace). Data is retained for 18 months at no extra charge.
> 
> So we are only ingesting each job time series for 5 min and then we stop exporting that set of jobs and then those time series become stale and are not ingested anymore. but customers are paying for the querying which they are technically paying for already with mysql.
> 
> Active time series are not growing "unbounded" here because we are always bounded by number of jobs that finished in the sacct query window. However for clusters with high activity this can become problematic. Steady-state active series = (job completion rate per hour) × (12 hour window) since active series are defined by # of time series that got ingested in the last 12 hours. but if we compare this to the # of metrics that get exported per cpu for our gpu skus the total number of jobs that finished in 12 hours may be miniscule.

See comment above. I am not sure you are understand how your records are being stored in the backend. The backend does not erase the data you are sending every 5 minutes. By backend I mean azure monitor workspace. (not your exporter).

PR #511, opened 2026-05-27
Add sdiag metrics to azslurm-exporter

azreenz:
Export metrics from sdiag in azslurm-exporter to get scheduler health overiview

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,7 +437,7 @@ To check if the configured exporters are exposing metrics, connect to a node and
 
 #### AzSlurm Exporter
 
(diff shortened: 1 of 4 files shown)


azreenz:
Corresponding sdiag Panel 
image

aditigaur4:
What is the status of this PR? Is this ready for review or not?

aditigaur4:
The PR looks good but make sure to get feedback on the Panel. I think for these metrics you want a time series panel. People want to be able see scheduler instability over time.  Certain metrics represent scheduler load and the fact that slurm might be processing work more slowly. That needs to be correlated with time.

aditigaur4 (approved):

PR #515, opened 2026-06-03
Add slurm.primary_scheduler_name to match secondary_scheduler_name

ryanhamel:
If a user has a custom name for the scheduler node, then there is no configuration to tell the installib what the primary scheduler node name is, like we do for the secondary scheduler node.


aditigaur4:
This PR looks good, is there a reason for the do-not-merge label?

aditigaur4 (approved):

PR #516, opened 2026-06-03
Move job accounting warning to BooleanCheckBox

gingi:


dpwatrous:
This looks good to me, but we should wait to merge until the 8.9.1 release is ready which supports the new widget attributes.

PR #525, opened 2026-06-22
Resctrict autoscale.json file permissions 

azreenz:
Problem:
The autoscale.json file containing CycleCloud credentials (username, password, URL) is written with world-readable permissions (0644) by default, allowing any local non-root user to read and impersonate the Slurm scheduler to CycleCloud. 
https://portal.microsofticm.com/imp/v5/incidents/details/31000000645885/msrc

Solution:
Write credentials to a secure temporary file with umask 077 (owner-only)
Use install -m 600 to atomically publish the final file as read/write only by root
Add cleanup trap to ensure temp files are always removed


ryanhamel (approved):

aditigaur4:
This PR should not have been merged.

PR #543, opened 2026-08-18
Add mysql datasource panels to azslurm dashboard and update panel plugin versions

azreenz:
Add Job History and Current Running Jobs panels to azslurm dashboards using mysql server data source. Update plugin versions of exisiting azslurm dashbaords


azreenz:
How can we include mysql datasource panels only when mysql datasource in configured?

